Issue #87, opened 2023-08-10
ObservationalModel Getter has only access to dataset


NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 10, 2023, 15:22

Would it be possible to give access to the `state` to the observational model? If I am not mistaken it contains variables of `dataset` but many others in addition. As describe in #84 I would need to access to a linked variable

NicolasGensollen:
In GitLab by @JulietteOrtholand on Nov 28, 2023, 16:27

It was linked to a misunderstanding

Issue #84, opened 2023-07-20
Add Joint model in refacto


NicolasGensollen:
In GitLab by @JulietteOrtholand on Jul 20, 2023, 17:06

I open the issue to start the reflection on the integration of the Joint model on the code refacto as I will only open an MR after !101 will be merged. 

I started from the I made diagram of Etienne refacto to see what I will need to add. Note that it is not perfectly up to date an might need some changes but it a good start.

![Joint1.drawio__1_](https://my-gitlab-bucket.s3.amazonaws.com/938594763/31fa0329592858dac9cf8150709af0907988ffbb4e49de3e84068add40c53347/Joint1.drawio__1_.png)

So next steps are:
- Review and merge !101,
- Add events in data existing data structure. We will only consider right censored events for now. Thus we need to store 2 types of information: the time of the event and whether it is censored or not. There is one piece of information per patient (as covariates) and they are stored in two distinct columns [EXAMPLE ?].
     - __Done:__ create attributes that are easy to access, such as time points, values, and masks: `event_time` (float) and `event_bool` (bool). In dataset, create `_construct_values(self, data: Data)` method to fill the new attributes. Update to_pandas method in dataset. Update individual_data.
     - __To do:__ Check that data `from_`and `to_` methods are compatible with additional event columns,
- Understand how Joint models (univ and multi) should interact with the model structure,
- Add Weibull distribution and observation classes,
- Add Joint model class.

NicolasGensollen:
In GitLab by @JulietteOrtholand on Jul 28, 2023, 18:30

I have design the DAG of my model:
![joint_dag.drawio](https://my-gitlab-bucket.s3.amazonaws.com/938594763/b3e28cdd5b2f3df6652982c9f0a8e371263e89f35975cb7d239241e844c6560e/joint_dag.drawio.png)

Next step implementation !

NicolasGensollen:
In GitLab by @JulietteOrtholand on Oct 5, 2023, 13:44

I have tried to go through the same process for my joint model as for the refacto branch. I have compared to dev to get an order of the error I could expect. The results are encouraging! results were produced with the branch fix_and_joint.

# Longitudinal parameters

[...]

PR #227, opened 2025-02-25
Enable the getter of observational models to use the state instead of dataset

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 22, 2023, 09:42

The observational model could only access variable that were stored in dataset. But for the joint model, I would need it to use a linked variable which is just a reparametrisation of a dataset variable, but it seems to me that it was cleaner to use a linked variable rather than make the computations inside of the observational model. Thus I needed the observational model to get access to the state instead of just the dataset. This MR is linked to the issue #87 and more broadly to #84 .

### DIFF
--- a/leaspy/models/abstract_model.py
+++ b/leaspy/models/abstract_model.py
@@ -92,16 +92,16 @@ class AbstractModel(BaseModel):
     """
 
     def __init__(
-        self,
-        name: str,
-        *,
-        # TODO? if we'd allow to pass a state there should be a all bunch of checks I guess? only "equality" of DAG is OK?
-        # (WIP: cf. comment regarding inclusion of state here)
-        # state: Optional[State] = None,
-        # TODO? Factory of `ObservationModel` instead? (typically one would need the dimension to instantiate the `noise_std` variable of the right shape...)
-        obs_models: Union[ObservationModel, Iterable[ObservationModel]],
-        fit_metrics: Optional[Dict[str, float]] = None,
-        **kwargs
+            self,
+            name: str,
+            *,
+            # TODO? if we'd allow to pass a state there should be a all bunch of checks I guess? only "equality" of DAG is OK?
+            # (WIP: cf. comment regarding inclusion of state here)
+            # state: Optional[State] = None,
+            # TODO? Factory of `ObservationModel` instead? (typically one would need the dimension to instantiate the `noise_std` variable of the right shape...)
+            obs_models: Union[ObservationModel, Iterable[ObservationModel]],
+            fit_metrics: Optional[Dict[str, float]] = None,
+            **kwargs
     ):
         super().__init__(name, **kwargs)
 
@@ -134,7 +134,7 @@ class AbstractModel(BaseModel):
     #     self.check_noise_model_compatibility(noise_model)
     #     self._noise_model = noise_model
     #
-    #def check_noise_model_compatibility(self, model: BaseNoiseModel) -> None:
+    # def check_noise_model_compatibility(self, model: BaseNoiseModel) -> None:
     #    """
     #    Raise a LeaspyModelInputError is the provided noise model isn't compatible with the model instance.
     #    This needs to be implemented in subclasses.
@@ -267,9 +267,10 @@ class AbstractModel(BaseModel):
         extra_vars = set(parameters).difference(self.dag)
         if len(extra_vars):
             raise LeaspyModelInputError(f"Unknown model variables: {extra_vars}")
+
         # TODO: check no DataVariable provided???
-        #extra_params = set(parameters).difference(cur_params)
-        #if len(extra_params):
+        # extra_params = set(parameters).difference(cur_params)
+        # if len(extra_params):
         #    # e.g. mixing matrix, which is a derived variable - checking their values only
         #    warnings.warn(f"Ignoring some provided values that are not model parameters: {extra_params}")
 
@@ -386,20 +387,20 @@ class AbstractModel(BaseModel):
 
         # Model supports and needs sources?
         has_sources = (
-            hasattr(self, 'source_dimension')
-            and isinstance(self.source_dimension, int)
-            and self.source_dimension > 0
+                hasattr(self, 'source_dimension')
+                and isinstance(self.source_dimension, int)
+                and self.source_dimension > 0
         )
 
         # Check parameters names
-        expected_parameters = set(['xi', 'tau'] + int(has_sources)*['sources'])
+        expected_parameters = set(['xi', 'tau'] + int(has_sources) * ['sources'])
         given_parameters = set(ips.keys())
         symmetric_diff = expected_parameters.symmetric_difference(given_parameters)
         if len(symmetric_diff) > 0:
             raise LeaspyIndividualParamsInputError(
-                    f'Individual parameters dict provided {given_parameters} '
-                    f'is not compatible for {self.name} model. '
-                    f'The expected individual parameters are {expected_parameters}.')
+                f'Individual parameters dict provided {given_parameters} '
+                f'is not compatible for {self.name} model. '
+                f'The expected individual parameters are {expected_parameters}.')
 
         # Check number of individuals present (with low constraints on shapes)
         ips_is_array_like = {k: is_array_like(v) for k, v in ips.items()}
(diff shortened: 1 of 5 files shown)


NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 22, 2023, 09:43

changed target branch from `master` to `v2`

NicolasGensollen:
In GitLab by @NicolasGensollen on Aug 22, 2023, 11:36

Commented on [leaspy/models/abstract_model.py line 977](https://github.com/aramis-lab/leaspy/compare/96b2a0346a4d9695962e510686f544767b2737f0..8052f376220460ed0ff6323970d32e1ee444e6ba#diff-196e847e3c942e778868226a64191c31R977)



It will be great to have the same API for these functions, i.e. `self._put_data_XXX(state, dataset)`

Could you update `_put_data_timepoints` to use the same signature (basically move the `WeightedTensor` creation inside the method) ?

NicolasGensollen:
In GitLab by @NicolasGensollen on Aug 22, 2023, 11:36

Commented on [leaspy/models/obs_models/_bernoulli.py line 26](https://github.com/aramis-lab/leaspy/compare/96b2a0346a4d9695962e510686f544767b2737f0..8052f376220460ed0ff6323970d32e1ee444e6ba#diff-c0f186a8da4bb6627c1bf03394cece42L26)



I would put these checks in the `_put_data_y_values` method at the `WeightedTensor` instantiation.

This method would then be a trivial `return state['y']` 

It seems like these Observation models are doing less and less work as we progress... :thinking:

NicolasGensollen:
In GitLab by @NicolasGensollen on Aug 22, 2023, 11:36

Commented on [leaspy/models/abstract_model.py line 976](https://github.com/aramis-lab/leaspy/compare/96b2a0346a4d9695962e510686f544767b2737f0..fd681457cf1e8f10d8dce3b65d1626091016b875#diff-196e847e3c942e778868226a64191c31R976)



Wouldn't this add "y" to the state again ?

NicolasGensollen:
In GitLab by @NicolasGensollen on Aug 22, 2023, 11:36

Thanks @JulietteOrtholand ! 

The MR LGTM, although I'm a bit preoccupied that doing this seems to bypass almost totally the need for the observation models. If we go down this road, you'll need to update the documentation of the base Observation model also.

We can discuss this later today.

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 09:19

added 1 commit

<ul><li>ded6df87 - Correction after Nicolas&#39; review</li></ul>

[Compare with previous version](/icm-institute/aramislab/leaspy/-/merge_requests/105/diffs?diff_id=767987398&start_sha=fd681457cf1e8f10d8dce3b65d1626091016b875)

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 09:27

added 1 commit

<ul><li>8052f376 - PEP8 format for modified files</li></ul>

[Compare with previous version](/icm-institute/aramislab/leaspy/-/merge_requests/105/diffs?diff_id=767996605&start_sha=ded6df8726bed3589cf827a8deeb3675aa04e184)

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 09:29

Commented on [leaspy/models/abstract_model.py line 977](https://github.com/aramis-lab/leaspy/compare/96b2a0346a4d9695962e510686f544767b2737f0..8052f376220460ed0ff6323970d32e1ee444e6ba#diff-196e847e3c942e778868226a64191c31R977)



I moved the `WeightedTensor` creation outside because I could not move the  `WeightedTensor` creation inside of `_put_data_timepoints`.

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 09:29

Commented on [leaspy/models/obs_models/_bernoulli.py line 26](https://github.com/aramis-lab/leaspy/compare/96b2a0346a4d9695962e510686f544767b2737f0..8052f376220460ed0ff6323970d32e1ee444e6ba#diff-c0f186a8da4bb6627c1bf03394cece42L26)



Done

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 09:29

Commented on [leaspy/models/abstract_model.py line 976](https://github.com/aramis-lab/leaspy/compare/96b2a0346a4d9695962e510686f544767b2737f0..fd681457cf1e8f10d8dce3b65d1626091016b875#diff-196e847e3c942e778868226a64191c31R976)



You are right!

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 09:29

Thanks for your review. I have made 2 commits, one for the changes I made after your remarks and one to format all the files I have changed into PEP8 format, I thought it would be easier to review. All the tests passed except `test_fi_logistic_binary` and `test_usecase_logistic_binary` which make numeric comparisons.

NicolasGensollen:
In GitLab by @NicolasGensollen on Aug 23, 2023, 13:30

Thanks for taking my comments into account @JulietteOrtholand ! 

I'm still having an internal debate about whether we should go in this direction or not.

My understanding of the `ObservationModel` class's responsibility is that it is supposed to provide to the `Model`  the variables related to observations and their attachment to the model (and the model then incorporates them in its `State`). In this MR, the `Model` class becomes responsible for this, at least partially.

I'm still having some trouble understanding precisely the reasons for which you want to do that, even after looking at issues #87 and #84. I suppose the diagram you linked contains the answer but it's not trivial for me to understand it with my limited knowledge of the model you are trying to implement.

If I understand correctly, you need to define a `LinkedVariable` which depends both on observations and other variables of the model's state. Because of this, you want to have an observation model for this variable but the getter method would need access to the state in order to have access to the other variables it depends on. Is that correct ?

Would it be possible to "split this variable in two" ? By that I mean having one observed variable with an associated observation model, and one or more LinkedVariables linking this observed variable with other variables in the state. Does this make sense ?

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 13:51

My issue is the following.

__Context:__
For a Weibull law with censure, parametrized with $(\nu, \rho)$,  the likelihood of the time at event $t_e$ is of the shape $L = S(t_e)*h(t_e)$, where $S(t) = \exp \left(- \left( \frac{e^{\xi_i}(t-\tau_i)}{\nu}\right)^{\rho}\right)$ and $h(t) = -\frac{S^{'}(t)}{S(t)} = \frac{\rho e^{\xi_i}}{\nu}  \left(\frac{e^{\xi_i} (t-\tau_i)}{\nu}\right)^{\rho-1}$. 

__With actual computation:__
If I want to compute the likelihood with the actual structure I need to take as input `xi`, `tau`, `rho`, `nu` as variables (even though a Weibull law should have only 2 parameters) and `time_event` as observed variable. 

__My idea:__
My idea was to use only two variables: `rho` and `nu*torch.exp(-xi)`, which corresponds to the variable of a Weibull law, and to apply that on the linked variable `time_event-tau`, so that the computations are in the DAG and the model and the observation model is generic. 

__Conclusion:__
If by "split this variable in two" you mean what I described in "With actual computation", then it would be possible. But I had the impression that the DAG was there to limit computations between all types of variables so that is why I wanted to change that. I hope it is more clear :smile:

NicolasGensollen:
In GitLab by @NicolasGensollen on Aug 23, 2023, 14:29

Thanks for the explanations @JulietteOrtholand it is definitely clearer now ! 

Yes, this is the general idea I had in mind when I wrote "split in two", thanks for writing this in a much more precise way.

I understand the concern of limiting computations, but here it comes at the cost of changing the program architecture, which is a pretty high cost even if we are in the process of refactoring things.

Moreover, I thought the point of having the DAG was to accurately represent the different variables of the model and the relationships between them. 

If the only reason to not go with the "actual computation" solution is related to computation concerns, I would encourage you to implement your model like that anyway, and see whether this cost is prohibitive or not. I think that it will be surprising if it would, but in this situation we could think at potential solutions then.

WDYT ?

NicolasGensollen:
In GitLab by @JulietteOrtholand on Aug 23, 2023, 15:32

Ok I understand. Even though the risk is to have one `ObservationModel` per `Model`, I guess it is a small cost even in a refractory time (otherwise you never stop ;)). Let's close this MR then.

Issue #102, opened 2024-03-11
Error introduced by event-as-weightedtensors-in-state


NicolasGensollen:
In GitLab by @JulietteOrtholand on Mar 11, 2024, 09:50

The merge of this MR introduced an error in the joint model. It seems that this is linked to the fact that `nll_attach_event_ind` is a weigthed tensor with the weights of `event`. It is an issue because the gibbs algorithm extract the weighted value, which does not make sense here and it forces some values to zero. 

I have not found yet how to correct that, but I'll continue to work on it this afternoon.

PR #247, opened 2025-02-25
Event tensors

NicolasGensollen:
In GitLab by @NicolasGensollen on Mar 25, 2024, 10:50

The goal of this MR is to find another way than !124 of fixing #102 

The issue with !124 is that it changes the return value type of method `_nll` from a `Tensor` to a `WeightedTensor` which is in contradiction with the base class and other subclasses.

Also, it changes the behavior of `_get_func_result_for_tensor_or_weighted_tensor` when the passed value is a `WeightedTensor` 

I think that the issue comes from the fact that we are using a `WeightedTensor` to model something that isn't really a weighted tensor in the same way as the other variables. In most cases, events should be handled in the same way as regular variables represented as weighted tensors, but in a few cases they should be handled differently because the weights have a slightly different meaning for events.

This MR proposes to introduce a new `EventTensor` class, which is a simple subclass of `WeightedTensor`. The main logic modifications are:

* The getter of `WeibullRightCensoredObservationModel` returns an `EventTensor` 
* All Weibull family objects expect an `EventTensor` for parameter `x` instead of a `WeightedTensor` 
* `_get_func_result_for_tensor_or_weighted_tensor` doesn't do anything with the weights when it receives an `EventTensor`

The MR also proposes to make the `_get_func()` method of the `SymbolicDistribution` class public since it seemed to simplify the code in the `WeibullRightCensoredObservationModel`

The rest of the modifications are basically PEP8 stuffs.

### DIFF
--- a/leaspy/models/obs_models/_weibull.py
+++ b/leaspy/models/obs_models/_weibull.py
@@ -1,9 +1,5 @@
-import torch
-
 from leaspy.variables.distributions import WeibullRightCensored
-from leaspy.variables.specs import VariableInterface
-from leaspy.utils.weighted_tensor import WeightedTensor
-from leaspy.utils.weighted_tensor import WeightedTensor, sum_dim
+from leaspy.utils.weighted_tensor import EventTensor
 from leaspy.io.data.dataset import Dataset
 
 from ._base import ObservationModel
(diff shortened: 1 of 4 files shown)


NicolasGensollen:
In GitLab by @NicolasGensollen on Mar 25, 2024, 12:07

added 5 commits

<ul><li>4edaec76...da1fc087 - 2 commits from branch <code>v2</code></li><li>f809616c - make _get_func a public method</li><li>f3104020 - implement EventTensor as simple subclass of WeightedTensor</li><li>669ca7d1 - use EventTensor in observation models</li></ul>

[Compare with previous version](/icm-institute/aramislab/leaspy/-/merge_requests/125/diffs?diff_id=962841385&start_sha=4edaec76c0de54f08ac71a3f7d30ad97dd2c045b)

NicolasGensollen:
In GitLab by @NicolasGensollen on Mar 25, 2024, 12:30

marked this merge request as **ready**

NicolasGensollen:
In GitLab by @NicolasGensollen on Mar 25, 2024, 12:30

requested review from @JulietteOrtholand

NicolasGensollen:
In GitLab by @NicolasGensollen on Mar 27, 2024, 13:37

Closing in favor of !128

PR #264, opened 2025-02-25
Draft: add likelihood in the graph

NicolasGensollen:
In GitLab by @JulietteOrtholand on Apr 10, 2024, 09:42

In the perspective of computing AIC, we will need to compute the likelihood. I am doing it for now outside of leaspy but adding it in the graph would simplify my computations.


NicolasGensollen:
In GitLab by @JulietteOrtholand on Apr 10, 2024, 10:07

marked this merge request as **draft**

PR #267, opened 2025-02-25
Fix the behaviour of realizations objects that were just reference to model parameters during first iteration

NicolasGensollen:
In GitLab by @little_nemo on Apr 23, 2024, 18:37

Fix https://gitlab.com/icm-institute/aramislab/leaspy/-/issues/109, by making sure that the `tensor` attribute of `PopulationRealization` objects initialized from model parameters are not just reference to the model parameter tensor. This caused the first iteration of the model to modify the value of the `model.parameters` tensors during sampling (and effectively made computation of regularization wrong — they should be computed based on the "frozen" version of model.parameters of previous iterations. This behaviour was only existent in the first iteration because the `compute_model_sufficient_statistics` method of `MultivariateModel` (and probably in other models as well, I have not checked) performs the following on the ambient realizations object, in which some correct "cloning" is performed.

```python
# unlink all sufficient statistics from updates in realizations!
realizations = realizations.clone()
```

FYI this behaviour is already fixed in v2, and was the cause of some discrepancies observed in the first iteration.

It might break a bunch of tests though, since computation might diverge at step 1 between current `dev` and this `fix`.


NicolasGensollen:
In GitLab by @little_nemo on Apr 24, 2024, 14:26

To Do:
- check that this fix now agrees with v2, and if yes, update the `_binary` test cases

NicolasGensollen:
In GitLab by @little_nemo on May 6, 2024, 16:01

added 1 commit

<ul><li>e90c37f7 - Update tests for the _binary testcases with the correct behaviour on first...</li></ul>

[Compare with previous version](/icm-institute/aramislab/leaspy/-/merge_requests/145/diffs?diff_id=1000795137&start_sha=628ddd028b3ba49147c7dd09dd0763e6d1a1961d)

NicolasGensollen:
In GitLab by @little_nemo on May 6, 2024, 18:56

added 1 commit

<ul><li>f6ec2198 - Update tests for the _binary testcases with the correct behaviour on first...</li></ul>

[Compare with previous version](/icm-institute/aramislab/leaspy/-/merge_requests/145/diffs?diff_id=1000989288&start_sha=e90c37f7dfb670ccc321d1cdcb6fcdf4a84321f6)

NicolasGensollen:
In GitLab by @little_nemo on May 6, 2024, 19:01

I could check that updating the values obtained for the three tests now broken in `dev` by the fix also allows them to pass correctly on v2. Hence I updated the values in f6ec2198aec22864105f894f0d625ab2a8eca23c. 

Note that I had to artificially specify the version to be something different from `2.0.0-dev` (in this case I chose to put `1.4.1`, but open for discussion as to what I should put here) in order for the new tests to be fixed in `v2` (the saving-format in `dev` is still different from `v2` so we still want the "quick n dirty" fixes used in the tests during loading of old models to apply, and this is decided based on the `leaspy_version` field of the models jsons, so we want this to be `<2.0.0`).

NicolasGensollen:
In GitLab by @little_nemo on May 7, 2024, 14:34

Okay now I don't really understand why the new tests are not working on the CI while they work locally, more digging to do I guess :sweat:

NicolasGensollen:
In GitLab by @little_nemo on Jul 30, 2024, 14:41

added 22 commits

<ul><li>f6ec2198...fe6bb61b - 20 commits from branch <code>dev</code></li><li>3f2af7e5 - Fix the behaviour of realizations objects that were just reference to model...</li><li>a5f60b3d - Update tests for the _binary testcases with the correct behaviour on first...</li></ul>

[Compare with previous version](/icm-institute/aramislab/leaspy/-/merge_requests/145/diffs?diff_id=1078397069&start_sha=f6ec2198aec22864105f894f0d625ab2a8eca23c)

NicolasGensollen:
In GitLab by @NicolasGensollen on Sep 3, 2024, 14:48

Closing as been solved by !167

PR #297, opened 2025-02-25
Change the source distribution

NicolasGensollen:
In GitLab by @JulietteOrtholand on Oct 25, 2024, 16:54

As explain in the slide below (the whole course is available at the end of the MR), the sources in an ICA cannot be gaussian... An for now in leaspy they are....
undefined

The idea was to replace the gaussian distribution by an other one. I first though about log-normal distribution, which is according to Stéphanie Allassonière paper (in ref at the end of the MR) "probably one of the most commonly used parametric models for ICA. One reason for this is that the logistic probability density function (p.d.f.) is easy to describe, smooth, with a shape similar to the Gaussian, but with heavier, exponential, tails.". But it is not eay to center it on 0 ... So I finally used a Laplace distribution.

### Where should the reviewer start ?
There are a few lines changed, and the code run. But still no scientific tests have been done. Thus, for now the main questions are:
- does the model still works: can it find back the model parameters (apart from the mixing matrix) and the random parameters,
- is now A identifiable?

## References
undefined

undefined

### DIFF
--- a/leaspy/models/abstract_multivariate_model.py
+++ b/leaspy/models/abstract_multivariate_model.py
@@ -19,7 +19,7 @@ from leaspy.variables.specs import (
     LinkedVariable,
     VariablesValuesRO,
 )
-from leaspy.variables.distributions import Normal
+from leaspy.variables.distributions import Laplace, Normal
 from leaspy.utils.functional import (
     Exp,
     MatMul,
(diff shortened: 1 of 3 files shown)


NicolasGensollen:
In GitLab by @JulietteOrtholand on Oct 25, 2024, 16:55

changed target branch from `master` to `v2`

NicolasGensollen:
In GitLab by @JulietteOrtholand on Feb 7, 2025, 17:59

This MR need to be tested, before going further. @sofia.kaisaridi are you still interested? Do you plan to test it? Or should I close it?

PR #300, opened 2025-02-25
the order of the dataset was not proper compare to tau

NicolasGensollen:
In GitLab by @JulietteOrtholand on Dec 5, 2024, 17:52

The order of the data object is not the same as the one of the tau object which imply some issues... I fix it.

### Where should the reviewer start ?

Only 2 lines

### How can the code be tested ?

I thought the test might fail due to the change in initialisation but no so ready to go !

### When is the MR due for? (review deadline)

ASAP

### What issues are linked to the MR ?

### DIFF
--- a/leaspy/models/joint.py
+++ b/leaspy/models/joint.py
@@ -222,7 +222,7 @@ class JointModel(LogisticMultivariateModel):
 
     def put_individual_parameters(self, state: State, dataset: Dataset):
 
-        df = dataset.to_pandas().reset_index('TIME').groupby('ID').min()
+        df = dataset.to_pandas().reset_index('TIME').groupby('ID').min().loc[dataset.indices]
 
         # Initialise individual parameters if they are not already initialised
         if not state.are_variables_set(('xi', 'tau')):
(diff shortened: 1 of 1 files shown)


NicolasGensollen:
In GitLab by @JulietteOrtholand on Dec 5, 2024, 18:05

requested review from @NicolasGensollen and @caglayantuna

NicolasGensollen:
In GitLab by @caglayantuna on Dec 6, 2024, 14:02

I couldn't understand where this order issue is coming from exactly. If there is an issue somewhere before, your change could be a temporary solution. Can you give more details? Maybe there is a bigger issue and we can fix it now

NicolasGensollen:
In GitLab by @JulietteOrtholand on Dec 20, 2024, 09:12

In `abstract_multivariate_model.py` there is a function `_get_dataframe_from_dataset`. I have made this MR because the index of my extracted dataframe was not in the same order as the one outputted by this function. But now I am wondering why is this function a method of the model and shouldn't it be in dataset class ? So we should not touch the index order of dataset inside of the model. What do you think ?

NicolasGensollen:
In GitLab by @JulietteOrtholand on Feb 7, 2025, 18:22

I close it, the !192 is now adressing the issue

Issue #117, opened 2024-12-05
What to do with individual parameters after fit


NicolasGensollen:
In GitLab by @JulietteOrtholand on Dec 5, 2024, 17:58

At the end of the fit, individual parameters are computed by the model. They correspond to the last iteration of the MCMC algorithm and users might want to have access to them event though they are only one draw and not really representative of the whole distribution. 

Maybe there is a reflexion to have around that in connexion with FitOutputManager and users needs.

NicolasGensollen:
In GitLab by @caglayantuna on Dec 12, 2024, 16:43

As I understand, `personalize` function with `scipy_miminize` returns an error when we use it with joint model, not for the others. It should be related to [this line](https://gitlab.com/icm-institute/aramislab/leaspy/-/blob/v2/src/leaspy/models/joint.py#L242). What is the target here? I am a little bit confused.

I don't know why, in abstract model, we have `put_individual_parameters` [here](https://gitlab.com/icm-institute/aramislab/leaspy/-/blob/v2/src/leaspy/models/abstract_model.py#L1002) but joint model has own `put_individual_parameters`. There is no `else` in abstract model which is very important difference between joint model. The problem occurs when we want to use this function for each individual separately. I believe that state.clone [here](https://gitlab.com/icm-institute/aramislab/leaspy/-/blob/v2/src/leaspy/algo/personalize/scipy_minimize.py#L606) is not correct because it clones every individuals.

I prefer to find a permanent solution when there is a bug because it seems that there are many operations can be simplified or removed. That's why I focused on personalize instead of fit.

NicolasGensollen:
In GitLab by @JulietteOrtholand on Dec 17, 2024, 10:41

I think that it returns an error with every type of model after a fit. For me the function `put_individual_parameters` is only used if the parameters are empty which is not the case for now after a fit, and thus it raise an error with `scipy_miminize`. 

But maybe we should check this is really the case with every model? Maybe what we want is to discard the individual parameters at the beginning of the `scipy_miminize` whatever situation. 

Fo the function `put_individual_parameters`, I think you are right there might be a bit of refactoring to do, but I am not sure if this is precisely related to this issue ? 

Let's discuss that this afternoon !

NicolasGensollen:
In GitLab by @laguilhon on Dec 18, 2024, 13:45

After testing it, there is this 'scipy_miminize' error only for 'univariate_joint' and 'multivariate' joint model.

[...]

PR #304, opened 2025-02-25
reset individual parameters

NicolasGensollen:
In GitLab by @JulietteOrtholand on Dec 19, 2024, 14:47

This is link to the issue #117


NicolasGensollen:
In GitLab by @JulietteOrtholand on Feb 7, 2025, 18:37

This might not be link to the original problem targeted and maybe a longer discussion is needed on that so #117 is there for that

Issue #123, opened 2025-02-20
Error: No file as Zeta.csv for joint multivariate competition


NicolasGensollen:
In GitLab by @laguilhon on Feb 20, 2025, 15:53

Error when using fit function for multivariate competing joint model:

`dataset = dataframe.set_index(["ID", "TIME"], verify_integrity=True).sort_index()`

`data_joint = Data.from_dataframe(dataset, "joint")`

`leaspy_joint = Leaspy("joint", nb_events=2, source_dimension=2)`

`algo_settings_joint_fit = AlgorithmSettings(`

`"mcmc_saem",`

`n_iter =1 000,`

`n_burn_in_iter= 500,`

`seed=0,`

`random_order_variables=True,`

` )`

`algo_settings_joint_fit.set_logs(`

`path="logs_multivariate_joint_",`

`plot_periodicity=50,`

`save_periodicity=1,`

`console_print_periodicity=None,`

`overwrite_logs_folder=True,`

` )`

`leaspy_joint.fit(data_joint, settings=algo_settings_joint_fit)`

> FileNotFoundError: \[Errno 2\] No such file or directory: '/Users/lea.aguilhon/Documents/CANAHIS/Code_Analyse/Joint_model/logs_multivariate_joint\_/parameter_convergence/zeta.csv'

Related to the new FOM functions that search file zeta.csv, for creating plots; but in competing model there is multi-zeta files that are created: zeta_0.csv, zeta_1.csv.

Zeta files need to be treated as for example mixing_matrix files, that works well.

Here is a dataset to test it: 

undefined

PR #327, opened 2025-02-28
improve pdf file created by FOM

caglayantuna:
This PR solves the issue #123 . 

To do this, I improved the `save_plot_convergence_model_parameters` function. Now, it creates a pdf with several pages (6 plots per page) and it plots some additional parameters such as mixing matrix for all models and zeta from joint model. Since these new parameters have multiple files, I have also added legends for them specifically. 

I add an example pdf which was created after fitting data with joint model.
[convergence_parameters.pdf](https://github.com/user-attachments/files/19029065/convergence_parameters.pdf)

### DIFF
--- a/src/leaspy/io/logs/fit_output_manager.py
+++ b/src/leaspy/io/logs/fit_output_manager.py
@@ -291,9 +291,9 @@ class FitOutputManager:
         colors = colormaps["Dark2"](np.linspace(0, 1, number_of_patient_plot + 2))
 
         fig, ax = plt.subplots(1, 1)
-        ax.set_title(f"Feature trajectory for {number_of_patient_plot} patients")
+        ax.set_title(f"Feature" f" trajectory for {number_of_patient_plot} patients")
         ax.set_xlabel("Ages")
-        ax.set_ylabel("Normalized Feature Value")
+        ax.set_ylabel("Normalized " "Feature Value")
 
         for i in range(number_of_patient_plot):
             times_patient = data.get_times_patient(i).cpu().detach().numpy()
--- a/src/leaspy/io/logs/fit_output_manager.py
+++ b/src/leaspy/io/logs/fit_output_manager.py
@@ -1,8 +1,10 @@
 import csv
 import math
 import os
+import re
 import time
 from pathlib import Path
+from typing import Iterable, Optional
 
 import matplotlib.cm as cm
 import matplotlib.pyplot as plt
@@ -185,12 +187,7 @@ class FitOutputManager:
 
         params_to_plot = list(model.state.tracked_variables - to_skip)
 
-        files = os.listdir(self.path_save_model_parameters_convergence)
-        files_to_plot = [
-            file
-            for file in files
-            if any(file.startswith(param) for param in params_to_plot)
-        ]
+        files_to_plot = self._get_files_related_to_parameters(params_to_plot)
         # To plot related parameters close to each other, we sort the list
         files_to_plot.sort()
 
@@ -201,9 +198,7 @@ class FitOutputManager:
         if self.plot_sourcewise:
             new_files = []
             for param_name in params_with_sources:
-                related_files = [
-                    file for file in files_to_plot if file.startswith(param_name)
-                ]
+                related_files = self._get_files_related_to_parameters([param_name])
                 if not related_files:
                     continue
                 related_files.sort()
@@ -213,25 +208,24 @@ class FitOutputManager:
                 for source_idx in range(num_sources):
                     combined_data = []
 
-                    for file_name in related_files:
-                        file_path = (
-                            self.path_save_model_parameters_convergence / file_name
-                        )
+                    for file_path in related_files:
                         df = pd.read_csv(file_path, index_col=0, header=None)
 
                         combined_data.append(df.iloc[:, source_idx])
 
                     combined_df = pd.concat(combined_data, axis=1, join="inner")
-                    new_file_name = "sourcewise_" f"{param_name}_{source_idx + 1}.csv"
+                    new_file_name = f"sourcewise_{param_name}_{source_idx + 1}.csv"
                     combined_df.to_csv(
                         self.path_save_model_parameters_convergence / new_file_name,
                         header=False,
                     )
-                    new_files.append(new_file_name)
+                    new_files.append(
+                        self.path_save_model_parameters_convergence / new_file_name
+                    )
             files_to_plot = [
                 file
                 for file in files_to_plot
-                if not any(param in file for param in params_with_sources)
+                if not any(file.name.startswith(param) for param in params_with_sources)
             ]
             files_to_plot.extend(new_files)
 
@@ -243,60 +237,27 @@ class FitOutputManager:
                 _, ax = plt.subplots(3, 2, figsize=(width, 3 * height_per_row))
                 ax = ax.flatten()
 
-                for i, file_name in enumerate(files_to_plot[page : page + 6]):
-                    file_path = self.path_save_model_parameters_convergence / file_name
-                    parameter_name = file_name.split(".csv")[0]
-
-                    if parameter_name != "v0":
-                        if parameter_name[-1].isdigit():
-                            # if there are more than 10 files (ex: mixing_matrix_10)
-                            if parameter_name[-2].isdigit():
-                                feature_index = int(parameter_name[-2:])
-                                parameter_name = parameter_name[:-3]
-                            else:
-                                feature_index = int(parameter_name[-1])
-                                parameter_name = parameter_name[:-2]
-
+                for i, file_path in enumerate(files_to_plot[page : page + 6]):
+                    parameter_name = file_path.name.split(".csv")[0]
+                    parameter_name, index = self._extract_parameter_name_and_index(
+                        parameter_name
+                    )
                     df_convergence = pd.read_csv(file_path, index_col=0, header=None)
                     ax[i].plot(df_convergence)
 
-                    if parameter_name == "mixing_matrix":
-                        ax[i].set_title(
-                            parameter_name + " " + model.features[feature_index]
-                        )
-                    elif parameter_name == "zeta":
-                        ax[i].set_title(
-                            parameter_name
-                            + " "
-                            + "event"
-                            + " "
-                            + str(feature_index + 1)
-                        )
-                    elif parameter_name.startswith("sourcewise"):
-                        ax[i].set_title(
-                            parameter_name.replace("sourcewise_", "")
-                            + " "
-                            + "source"
-                            + " "
-                            + str(feature_index)
-                        )
-                        ax[i].legend(model.features, loc="best")
-                    else:
-                        ax[i].set_title(parameter_name)
+                    ax = self._set_title_for_parameter(
+                        ax, i, parameter_name, model, index
+                    )
 
-                    if parameter_name in params_with_feature_labels:
-                        ax[i].legend(model.features, loc="best")
-                    if parameter_name in params_with_sources:
-                        sources = [
-                            "Source" + " " + str(i + 1)
-                            for i in range(model.source_dimension)
-                        ]
-                        ax[i].legend(sources, loc="best")
-                    if parameter_name in params_with_events:
-                        events = [
-                            "Event" + " " + str(i + 1) for i in range(model.nb_events)
-                        ]
-                        ax[i].legend(events, loc="best")
+                    ax = self._set_legend_for_parameters(
+                        ax,
+                        i,
+                        parameter_name,
+                        params_with_feature_labels,
+                        params_with_sources,
+                        params_with_events,
+                        model,
+                    )
 
                 plt.tight_layout()
                 pdf.savefig()
@@ -394,3 +355,66 @@ class FitOutputManager:
 
         plt.savefig(path_iteration)
         plt.close()
+
+    def _get_files_related_to_parameters(self, parameters: Iterable[str]) -> list[Path]:
+        return [
+            f
+            for f in self.path_save_model_parameters_convergence.iterdir()
+            if any(f.name.startswith(param) for param in parameters)
+        ]
+
+    def _extract_parameter_name_and_index(
+        self, parameter_name: str
+    ) -> tuple[Optional[str], Optional[int]]:
+        if parameter_name == "v0":
+            return parameter_name, None
+        match = re.search(r"^(.*?)(\d+)$", parameter_name)
+        if match:
+            return match.group(1).strip("_"), int(match.group(2))
+        else:
+            return parameter_name, None
+
+    def _set_title_for_parameter(self, ax, i, parameter_name: str, model, index):
+        if parameter_name == "mixing_matrix":
+            ax[i].set_title(parameter_name + " " + model.features[index])
+        elif parameter_name == "zeta":
+            ax[i].set_title(parameter_name + " " + "event" + " " + str(index + 1))
+        elif parameter_name.startswith("sourcewise"):
+            ax[i].set_title(
+                parameter_name.replace("sourcewise_", "")
+                + " "
+                + "source"
+                + " "
+                + str(index)
+            )
+        else:
+            ax[i].set_title(parameter_name)
+        return ax
+
+    def _set_legend_for_parameters(
+        self,
+        ax,
+        i,
+        parameter_name,
+        params_with_feature_labels,
+        params_with_sources,
+        params_with_events,
+        model,
+    ):
+        if parameter_name in params_with_feature_labels:
+            ax[i].legend(model.features, loc="best")
+        if parameter_name in params_with_sources:
+            sources = [
+                "Source" + " " + str(i + 1) for i in range(model.source_dimension)
+            ]
+            ax[i].legend(sources, loc="best")
+        if parameter_name in params_with_events:
+            events = ["Event" + " " + str(i + 1) for i in range(model.nb_events)]
+            ax[i].legend(events, loc="best")
+        if parameter_name.startswith("sourcewise"):
+            if "mixing_matrix" in parameter_name:
+                ax[i].legend(model.features, loc="best")
+            if "zeta" in parameter_name:
+                events = ["Event" + " " + str(i + 1) for i in range(model.nb_events)]
+                ax[i].legend(events, loc="best")
+        return ax
--- a/src/leaspy/io/logs/fit_output_manager.py
+++ b/src/leaspy/io/logs/fit_output_manager.py
@@ -196,38 +196,9 @@ class FitOutputManager:
 
         # If plot sourcewise is true, new sourcewise csv files will be created
         if self.plot_sourcewise:
-            new_files = []
-            for param_name in params_with_sources:
-                related_files = self._get_files_related_to_parameters([param_name])
-                if not related_files:
-                    continue
-                related_files.sort()
-
-                num_sources = model.source_dimension
-
-                for source_idx in range(num_sources):
-                    combined_data = []
-
-                    for file_path in related_files:
-                        df = pd.read_csv(file_path, index_col=0, header=None)
-
-                        combined_data.append(df.iloc[:, source_idx])
-
-                    combined_df = pd.concat(combined_data, axis=1, join="inner")
-                    new_file_name = f"sourcewise_{param_name}_{source_idx + 1}.csv"
-                    combined_df.to_csv(
-                        self.path_save_model_parameters_convergence / new_file_name,
-                        header=False,
-                    )
-                    new_files.append(
-                        self.path_save_model_parameters_convergence / new_file_name
-                    )
-            files_to_plot = [
-                file
-                for file in files_to_plot
-                if not any(file.name.startswith(param) for param in params_with_sources)
-            ]
-            files_to_plot.extend(new_files)
+            files_to_plot = self._update_files_sourcewise(
+                files_to_plot, params_with_sources, model.source_dimension
+            )
 
         n_plots = len(files_to_plot)
         n_rows = math.ceil(n_plots / 2)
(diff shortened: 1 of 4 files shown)


src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -169,31 +171,77 @@ def save_plot_convergence_model_parameters(self, model: AbstractModel):
...
+                            j += 1
+                        else:
+                            break
+
+                    file_path = (
+                        self.path_save_model_parameters_convergence
+                        / f"{parameter_name}.csv"
+                    )
+                    if os.path.exists(file_path):
+                        file_to_plot = file_path
```
NicolasGensollen, line 211: Lines 191 to 211 can be replaced by:

```python
files_to_plot = [
    f for f in self.path_save_model_parameters_convergence.glob(f"{parameter_name}*.csv")
]
```

If needed the regexp can be made more robust but I think this should be enough.

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -169,31 +171,77 @@ def save_plot_convergence_model_parameters(self, model: AbstractModel):
...
+
+                    if any(
+                        f"_{j}" in str(file) for j, file in enumerate(files_to_plot)
```
NicolasGensollen, line 214: My previous suggestion gives you a list of files matching your query. This list can be of length zero, one or more. I suggest you handle these three cases rather than this strange condition.

```python
# in this function you'd have:
for i, parameter_name in enumerate(params_to_plot[page : page + 6]):
    files_to_plot = [
        f for f in self.path_save_model_parameters_convergence.glob(f"{parameter_name}*.csv")
    ]
    _plot_parameter(files_to_plot, ax[i])

# with helper functions:
def _plot_parameter(files_to_plot: Sequence[Path], ax: Axes):
    if len(files_to_plot) == 0:
        # Do something...
    elif len(files_to_plot) == 1:
        _plot_single_parameter_file(files_to_plot[0], ax)
    else:
        _plot_multiple_parameter_file(files_to_plot, ax)

def _plot_single_parameter_file(file_to_plot: Path, ax: Axes) -> None:
    df = pd.read_csv(file_to_plot, index_col=0, header=None)
    ax.plot(df)

def _plot_multiple_parameter_file(files_to_plot: Iterable[Path], ax: Axes) -> None:
    colormap = cm.viridis(np.linspace(0, 1, len(files_to_plot)))
    for idx, file in enumerate(files_to_plot):
        df = pd.read_csv(file, index_col=0, header=None)
        line = ax.plot(df, color=colormap[idx])[0]
        lines.append(line)
    ax.legend(lines, [f.stem for f in files_to_plot], loc="best")
```

Does that make sense ?

NicolasGensollen (commented):
Thanks @caglayantuna !

I made some suggestion to improve the code readability.

Also, it would be nice to add some unit tests for these new functions.

caglayantuna:
Thanks @NicolasGensollen for your comments. I changed this PR after some discussions with the team. Review is not necessary for now. I am attaching latest pdf results to discuss about them.
[convergence_parameters_joint.pdf](https://github.com/user-attachments/files/19127948/convergence_parameters_joint.pdf)
[convergence_parameters_logistic.pdf](https://github.com/user-attachments/files/19127955/convergence_parameters_logistic.pdf)

caglayantuna:
This PR is ready for reviewing. I added a test function for `FitOutputManager`. This function tests some functions of `FitOutputManager` class except print functions. I couldn't find how to test print based functions. 
 
Also, I added `plot_sourcewise` option to logs. I share 2 pdf results when `plot_sourcewise` is true.

[convergence_parameters_joint_plot_sourcewise.pdf](https://github.com/user-attachments/files/19367812/convergence_parameters_joint_plot_sourcewise.pdf)


[convergence_parameters_logistic_plot_sourcewise.pdf](https://github.com/user-attachments/files/19367814/convergence_parameters_logistic_plot_sourcewise.pdf)

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+        params_with_sources = ["mixing_matrix"]
+        params_with_events = []
+        if model.name == "joint":
+            to_skip.add("survival_shifts")
+            params_with_sources.append("zeta")
+            params_with_events += ["nu", "rho"]
+
+        params_to_plot = list(model.state.tracked_variables - to_skip)
+
+        files = os.listdir(self.path_save_model_parameters_convergence)
```
NicolasGensollen, line 188: `self.path_save_model_parameters_convergence` should be a Path object and not a string since it is a path.

Also, the logic used to get the files is duplicated, I'd turn it into a function:

```python
def _get_files_related_to_parameters(self, parameters: Iterable[str]) - > list[Path]:
    return [
        f for f in self.path_save_model_parameters_convergence.iterdir()
        if any(f.name.startswith(param) for param in parameters)
    ]
```

So getting the files to plot here would become:

```python
files_to_plot = self._get_files_related_to_parameters(params_to_plot).sort()
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
-            df_convergence.plot(ax=ax[x_position][y_position], legend=False)
-            ax[x_position][y_position].set_title(parameter_name)
-
-        plt.tight_layout()
-        plt.savefig(self.path_plot_convergence_model_parameters)
-        plt.close()
+        # If plot sourcewise is true, new sourcewise csv files will be created
+        if self.plot_sourcewise:
+            new_files = []
+            for param_name in params_with_sources:
```
NicolasGensollen, line 203: This would become:

```python
related_files = self._get_files_related_to_parameters(params_with_sources)
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+                if not related_files:
+                    continue
+                related_files.sort()
+
+                num_sources = model.source_dimension
+
+                for source_idx in range(num_sources):
+                    combined_data = []
+
+                    for file_name in related_files:
```
NicolasGensollen, line 216: This would also be simplified since the list holds the full path:

```python
for file_path in related_files:
    ...
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+                        file_path = (
+                            self.path_save_model_parameters_convergence / file_name
+                        )
+                        df = pd.read_csv(file_path, index_col=0, header=None)
+
+                        combined_data.append(df.iloc[:, source_idx])
+
+                    combined_df = pd.concat(combined_data, axis=1, join="inner")
+                    new_file_name = "sourcewise_" f"{param_name}_{source_idx + 1}.csv"
```
NicolasGensollen, line 225: This looks weird, is this what you wanted to write:

```suggestion
                    new_file_name = f"sourcewise_{param_name}_{source_idx + 1}.csv"
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+
+        n_plots = len(files_to_plot)
+        n_rows = math.ceil(n_plots / 2)
+        with PdfPages(self.path_plot_convergence_model_parameters) as pdf:
+            for page in range(0, n_plots, 6):
+                # 6 plots per page
+                _, ax = plt.subplots(3, 2, figsize=(width, 3 * height_per_row))
+                ax = ax.flatten()
+
+                for i, file_name in enumerate(files_to_plot[page : page + 6]):
```
NicolasGensollen, line 246:
```suggestion
                for i, file_path in enumerate(files_to_plot[page : page + 6]):
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+                    file_path = self.path_save_model_parameters_convergence / file_name
```
NicolasGensollen, line 247:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+                    parameter_name = file_name.split(".csv")[0]
```
NicolasGensollen, line 248:
```suggestion
                    parameter_name = file_path.name.split(".csv")[0]
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+
+                    if parameter_name != "v0":
+                        if parameter_name[-1].isdigit():
+                            # if there are more than 10 files (ex: mixing_matrix_10)
+                            if parameter_name[-2].isdigit():
+                                feature_index = int(parameter_name[-2:])
+                                parameter_name = parameter_name[:-3]
+                            else:
+                                feature_index = int(parameter_name[-1])
+                                parameter_name = parameter_name[:-2]
```
NicolasGensollen, line 258: This is very fragile and not generalizable. You need to use regular expressions for these kind of tasks.

Here is a proposition which could still be improved:

```python
def _extract_parameter_name_and_index(parameter_name: str) -> tuple[Optional[str], Optional[int]]:
    match = re.search(r'^(.*?)(\d+)$', parameter_name)
    if match:
        return match.group(1).strip("_"), int(match.group(2))
    return None, None
```

Which gives:

```python
>>> _extract_parameter_name_and_index("mixing_matrix_10")
('mixing_matrix', 10)
>>> _extract_parameter_name_and_index("mixing_matrix_1")
('mixing_matrix', 1)
>>> _extract_parameter_name_and_index("mixing_matrix_10400")
('mixing_matrix', 10400)
>>> _extract_parameter_name_and_index("mixing_matrix_10400_foo")
(None, None)
>>> _extract_parameter_name_and_index("mixing_ma")
(None, None)
>>> _extract_parameter_name_and_index("100_mixing_ma")
(None, None)
>>> _extract_parameter_name_and_index("v0_12")
('v0', 12)
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -159,41 +162,145 @@ def save_model_parameters_convergence(
...
+
+                    df_convergence = pd.read_csv(file_path, index_col=0, header=None)
+                    ax[i].plot(df_convergence)
+
+                    if parameter_name == "mixing_matrix":
```
NicolasGensollen, line 263: This looks very hardcoded, would there be a way to have something more generic ?

Also, could you move that to a separate function, this function is enormous and hard to follow along. Something like this maybe:

```python
def _set_title_for_parameter(ax, parameter_name: str):
    if parameter_name == "mixing_matrix":
        ax.set_title(...)
    ...
```

NicolasGensollen (commented):
Thanks @caglayantuna 

I made a first pass. My general comment is that the function you added is very very long, and hence difficult to read and understand as a whole.
I think the code would read much better by extracting some logic in smaller functions.

src/leaspy/io/logs/fit_output_manager.py
caglayantuna: done

src/leaspy/io/logs/fit_output_manager.py
caglayantuna: thanks, since I added `sourcewise` later, I forgot to change "f". now , it is ok

src/leaspy/io/logs/fit_output_manager.py
caglayantuna: Thanks, done

src/leaspy/io/logs/fit_output_manager.py
caglayantuna: I removed it

src/leaspy/io/logs/fit_output_manager.py
caglayantuna: Thank you very much for your suggestions, it is very useful. However, I couldn't find a solution when input is `v0`. This is a parameter name and `0` is not index. To fix this, I had to add specific condition for this parameter. If you have any suggestion, would be nice.

src/leaspy/io/logs/fit_output_manager.py
caglayantuna: Normally, we could use `params_with_sources` instead of hardcoding. However, `zeta` in joint model has different structure. I couldn't find a better solution but I created a function as you suggested.

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
 
         plt.savefig(path_iteration)
         plt.close()
+
+    def _get_files_related_to_parameters(self, parameters: Iterable[str]) -> list[Path]:
+        """
+        Helper function for `save_plot_convergence_model_parameters`
```
NicolasGensollen, line 332:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+
```
NicolasGensollen, line 333:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+            f
+            for f in self.path_save_model_parameters_convergence.iterdir()
+            if any(f.name.startswith(param) for param in parameters)
+        ]
+
+    def _extract_parameter_name_and_index(
+        self, parameter_name: str
+    ) -> tuple[Optional[str], Optional[int]]:
+        """
+        Helper function for `save_plot_convergence_model_parameters`
```
NicolasGensollen, line 356:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+
```
NicolasGensollen, line 357:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+        Parameters
+        ----------
+        parameter_name : :obj:`str`
+            The name of the parameter to extract information from.
+
+        Returns
+        -------
+        tuple[Optional[:obj:`str`], Optional[:obj:`int`]]
+            A tuple where the first element is the parameter name and the second is its index (if applicable).
```
NicolasGensollen, line 368: Maybe a small examples section could help the reader understand very quickly what this function is doing, WDYT ?

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+            return parameter_name, None
+        match = re.search(r"^(.*?)(\d+)$", parameter_name)
+        if match:
+            return match.group(1).strip("_"), int(match.group(2))
+        else:
+            return parameter_name, None
+
+    def _set_title_for_parameter(self, ax, i, parameter_name: str, model, index):
+        """
+        Helper function for `save_plot_convergence_model_parameters`
```
NicolasGensollen, line 380:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+
```
NicolasGensollen, line 381:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+        """
+        if parameter_name == "v0":
+            return parameter_name, None
+        match = re.search(r"^(.*?)(\d+)$", parameter_name)
+        if match:
+            return match.group(1).strip("_"), int(match.group(2))
+        else:
+            return parameter_name, None
+
+    def _set_title_for_parameter(self, ax, i, parameter_name: str, model, index):
```
NicolasGensollen, line 378: Please add the type hints here.

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+
+    def _set_legend_for_parameters(
+        self,
+        ax,
+        i,
+        parameter_name,
+        params_with_feature_labels,
+        params_with_sources,
+        params_with_events,
+        model,
```
NicolasGensollen, line 426: Please add type hints here as well.

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+    ):
+        """
+        Helper function for `save_plot_convergence_model_parameters`
```
NicolasGensollen, line 429:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+
```
NicolasGensollen, line 430:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+            ax[i].legend(events, loc="best")
+        if parameter_name.startswith("sourcewise"):
+            if "mixing_matrix" in parameter_name:
+                ax[i].legend(model.features, loc="best")
+            if "zeta" in parameter_name:
+                events = ["Event" + " " + str(i + 1) for i in range(model.nb_events)]
+                ax[i].legend(events, loc="best")
+        return ax
+
+    def _update_files_sourcewise(self, files_to_plot, params_with_sources, num_sources):
```
NicolasGensollen, line 473: type hints

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+        """
+        Helper function for `save_plot_convergence_model_parameters`
```
NicolasGensollen, line 475:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+
```
NicolasGensollen, line 476:
```suggestion
```

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+        Update the list of files by creating sourcewise files based on the parameters with sources.
+
+        This function processes parameters that have sources and generates new CSV files for each source.
+        These new files are added to the list of files to plot.
```
NicolasGensollen, line 480: Two small comments here:

First of all, this function is doing multiple things, and it's clearly visible when reading the description. Because of that, it's long, and not easy to understand.

Second, it is mutating the `files_to_plot` list for no reason. A safer implementation would compute the `new_files` list and return that. The caller would then perform the aggregation of the `files_to_plot` and `new_files`. Doing that, this function doesn't even need the `files_to_plot` list. It could basically look like that:

```python
def _compute_files_sourcewise(self, params_with_sources: list[str], num_sources: int) -> list[Path]:
    new_files = []
    for param_name in params_with_sources:
        ...
    return new_files
```

The caller:

```python
files_to_plot = [
    file
    for file in files_to_plot
    if not any(file.name.startswith(param) for param in params_with_sources)
]
files_to_plot.extend(self._compute_files_sourcewise(...))
```

WDYT ?

src/leaspy/io/logs/fit_output_manager.py
```diff
@@ -287,3 +326,201 @@ def save_plot_patient_reconstructions(
...
+        for param_name in params_with_sources:
+            related_files = self._get_files_related_to_parameters([param_name])
+            if not related_files:
+                continue
+            related_files.sort()
+
+            for source_idx in range(num_sources):
+                combined_data = []
+
+                for file_path in related_files:
```
NicolasGensollen, line 506: We have three imbricated for loop in this function. It's another sign that we are trying to do too much at once.

For each source, we read the related files one by one and concatenate everything together, this concatenation is then saved to disk and the path is added to the `new_files` list.

Here is a suggestion of how we could refactor this to help readability:

```python
...
for source_idx in range(num_sources):
    new_files.append(
        _concatenate_and_save_sourcewise_parameters(
            param_name, related_files, source_idx
        )
    )
...

def _concatenate_and_save_sourcewise_parameters(
    self, parameter_name: str, files: Iterable[Path], source_idx: int
) -> Path:
    """Explain what I do..."""
    df = _concatenate_parameters(files, source_idx)
    file_name = f"sourcewise_{parameter_name}_{source_idx + 1}.csv"
    file_path = self.path_save_model_parameters_convergence / file_name
    df.to_csv(file_path, header=False)
    return file_path

def _concatenate_parameters(files: Iterable[Path], source_idx: int) -> pd.DataFrame:
    """Explain me..."""
    return pd.concat(
        [
            pd.read_csv(file_path, index_col=0, header=None).iloc[:, source_idx]
            for file_path in files
        ],
        axis=1,
        join="inner",
    )
            
```

caglayantuna:
Hello @NicolasGensollen , thanks for your suggestions again. I changed the PR accordingly.

NicolasGensollen (approved):
LGTM, thanks @caglayantuna !

Issue #122, opened 2025-02-20
Clean docstrings of V2


NicolasGensollen:
## Objectives

As discussed in recent meetings, we need to take care of the docstrings of Leaspy V2. The goals are:

- make sure the docstrings exist (at least for public objects)
- make sure the docstrings are correctly formatted (Numpydoc style, cross-references etc...)
- make sure the texts of the docstrings are correct and complete (sometimes they are out-of-sync with the code, sometimes they don't make any sense, sometimes they are pretty obscure...)
- try to improve the docstrings as much as possible (add type hints, add references and so on)

## Style and references

In terms of style, we use the [Numpydoc style](https://numpydoc.readthedocs.io/en/latest/format.html) to write our docstrings, they should basically look like that:

```python
def do_something(parameter1: int, parameter2: str = "all") -> float:
    """One line short description of what the function does.

    Eventually a more detailed explanation which can span multiple lines.

    Parameters
    ----------
    parameter1 : :obj:`int`
        Description of parameter1.

    parameter2 : :obj:`str`, optional
        Description of parameter2.
        Default="all".

    Returns
    -------
    :obj:`float` :
        Description of the return value.
    """
    ....
```

In docstrings it is possible to link other functions or classes of Leaspy by using the following syntax: `:class:~leaspy.models.AbstractModel`. You can take a look at MR !189 to see some examples.

## Work organization

As discussed, we can share the load between us. I'm going to propose a split but we can re-organize things if some of you want to do more/less.

The idea would be to open small MR targeting a small number of files at the time. We could for example have one MR per bullet point if that makes sense. 

### `leaspy.algo`

- [ ] `leaspy.algo.fit` : @caglayantuna 
- [ ] `leaspy.algo.personalize` : @caglayantuna 
- [ ] `leaspy.algo.others` : @caglayantuna 
- [ ] `leaspy.algo.utils` : @caglayantuna 
- [ ] `leaspy.algo.base.py` : @caglayantuna 
- [ ] `leaspy.algo.factory.py` : @caglayantuna 
- [ ] `leaspy.algo.settings.py` : @caglayantuna 

### `leaspy.datasets`

- [ ] `leaspy.datasets` : @NicolasGensollen (It's more or less already done, I'll finish it if needed...)

### `leaspy.io`

- [ ] `leaspy.io.data.data.py` : @laguilhon
- [ ] `leaspy.io.data.dataset.py` : @maylistran01 
[...]

PR #330, opened 2025-03-05
clean docstrings of algo base

caglayantuna:
This PR updates docstrings of algo/base related to issue #122 .

### DIFF
--- a/src/leaspy/algo/base.py
+++ b/src/leaspy/algo/base.py
@@ -1,3 +1,5 @@
+"""This module defines the `AlgorithmType`, `AlgorithmName` and `AbstractAlgo` classes"""
+
 import random
 import sys
 import time
@@ -26,6 +28,8 @@ __all__ = [
 
 
 class AlgorithmType(str, Enum):
+    """The type of the algorithms."""
+
     FIT = "fit"
     PERSONALIZE = "personalize"
     SIMULATE = "simulate"
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
 class AbstractAlgo(ABC):
     """
     Abstract class containing common methods for all algorithm classes.
-    These classes are child classes of `AbstractAlgo`.
+    These classes are children of `AbstractAlgo`.
 
     Parameters
     ----------
-    settings : :class:`.AlgorithmSettings`
-        The specifications of the algorithm as a :class:`.AlgorithmSettings` instance.
+    settings : :class:`~leaspy.algo.AlgorithmSettings`
+        The specifications of the algorithm as a :class:`~leaspy.algo.AlgorithmSettings` instance.
 
     Attributes
     ----------
-    name : str
+    name : :obj:`str`
         Name of the algorithm.
-    family : str
+    family : :obj:`str`
         Family of the algorithm. For now, valid families are:
-            * ``'fit'```
-            * ``'personalize'```
+            * ``'fit'``
+            * ``'personalize'``
             * ``'simulate'``
-    deterministic : bool
+    deterministic : :obj:`bool`
         True, if and only if algorithm does not involve in randomness.
-        Setting a seed and such algorithms will be useless.
-    algo_parameters : dict
+        Setting a seed will have no effect on such algorithms.
+    algo_parameters : :obj:`dict`
         Contains the algorithm's parameters. Those are controlled by
-        the :attr:`.AlgorithmSettings.parameters` class attribute.
-    seed : int, optional
+        the :attr:`leaspy.algo..AlgorithmSettings.parameters` class attribute.
+    seed : :obj:`int`, optional
         Seed used by :mod:`numpy` and :mod:`torch`.
-    output_manager : :class:`~.io.logs.fit_output_manager.FitOutputManager`
+    output_manager : :class:`~leaspy.io.logs.fit_output_manager.FitOutputManager`
         Optional output manager of the algorithm
     """
 
@@ -118,7 +122,7 @@ class AbstractAlgo(ABC):
         model: AbstractModel,
         *args,
         **extra_kwargs,
-    ) -> Tuple[Any, Optional[torch.FloatTensor]]:
+    ) -> tuple[Any, Optional[torch.Tensor]]:
         """
         Run the algorithm (actual implementation), to be implemented in children classes.
 
@@ -126,9 +130,9 @@ class AbstractAlgo(ABC):
 
         Parameters
         ----------
-        model : :class:`~.models.abstract_model.AbstractModel`
+        model : :class:`~leaspy.models.AbstractModel`
             The used model.
-        dataset : :class:`.Dataset`
+        dataset : :class:`~leaspy.io.data.Dataset`
             Contains all the subjects' observations with corresponding timepoints, in torch format to speed up computations.
 
         Returns
@@ -154,11 +158,11 @@ class AbstractAlgo(ABC):
 
         Parameters
         ----------
-        model : :class:`~.models.abstract_model.AbstractModel`
+        model : :class:`~leaspy.models.AbstractModel`
             The used model.
-        dataset : :class:`.Dataset`
+        dataset : :class:`~leaspy.io.data.Dataset`
             Contains all the subjects' observations with corresponding timepoints, in torch format to speed up computations.
-        return_loss : bool (default False), keyword only
+        return_loss : :obj:`bool` (default False), keyword only
             Should the algorithm return main output and optional loss output as a 2-tuple?
 
         Returns
@@ -201,7 +205,7 @@ class AbstractAlgo(ABC):
 
         Parameters
         ----------
-        parameters : dict
+        parameters : :obj:`dict`
             Contains the pairs (key, value) of the wanted parameters
 
         Examples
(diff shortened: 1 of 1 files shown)


src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
     ----------
-    settings : :class:`.AlgorithmSettings`
-        The specifications of the algorithm as a :class:`.AlgorithmSettings` instance.
+    settings : :class:`~leaspy.algo.AlgorithmSettings`
+        The specifications of the algorithm as a :class:`~leaspy.algo.AlgorithmSettings` instance.
 
     Attributes
     ----------
-    name : str
+    name : :obj:`str`
```
NicolasGensollen, line 63:
```suggestion
    name : :class:`~leaspy.algo.base.AlgorithmName`
```

src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
         Name of the algorithm.
-    family : str
+    family : :obj:`str`
```
NicolasGensollen, line 65:
```suggestion
    family : :class:`~leaspy.algo.base.AlgorithmType`
```

src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
         Family of the algorithm. For now, valid families are:
```
NicolasGensollen, line 66: We don't need to describe all possible values because the enumeration is doing it for us already:

```suggestion
        Family of the algorithm.
```

src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
-            * ``'fit'```
-            * ``'personalize'```
+            * ``'fit'``
```
NicolasGensollen, line 67:
```suggestion
```

src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
+            * ``'personalize'``
```
NicolasGensollen, line 68:
```suggestion
```

src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
             * ``'simulate'``
```
NicolasGensollen, line 69:
```suggestion
```

src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
-    deterministic : bool
+    deterministic : :obj:`bool`
         True, if and only if algorithm does not involve in randomness.
```
NicolasGensollen, line 71:
```suggestion
        True, if and only if algorithm does not involve randomness.
```

src/leaspy/algo/base.py
```diff
@@ -47,31 +51,31 @@ class AlgorithmName(str, Enum):
...
-        Setting a seed and such algorithms will be useless.
-    algo_parameters : dict
+        Setting a seed will have no effect on such algorithms.
+    algo_parameters : :obj:`dict`
         Contains the algorithm's parameters. Those are controlled by
-        the :attr:`.AlgorithmSettings.parameters` class attribute.
-    seed : int, optional
+        the :attr:`leaspy.algo..AlgorithmSettings.parameters` class attribute.
```
NicolasGensollen, line 75:
```suggestion
        the :attr:`leaspy.algo.AlgorithmSettings.parameters` class attribute.
```

src/leaspy/algo/base.py
```diff
@@ -118,17 +122,17 @@ def run_impl(
         model: AbstractModel,
         *args,
         **extra_kwargs,
-    ) -> Tuple[Any, Optional[torch.FloatTensor]]:
+    ) -> tuple[Any, Optional[torch.Tensor]]:
         """
         Run the algorithm (actual implementation), to be implemented in children classes.
 
         TODO fix proper abstract class
```
NicolasGensollen, line 129: Same comment:

```suggestion
```

src/leaspy/algo/base.py
```diff
@@ -154,11 +158,11 @@ def run(
...
         Parameters
         ----------
-        model : :class:`~.models.abstract_model.AbstractModel`
+        model : :class:`~leaspy.models.AbstractModel`
             The used model.
-        dataset : :class:`.Dataset`
+        dataset : :class:`~leaspy.io.data.Dataset`
             Contains all the subjects' observations with corresponding timepoints, in torch format to speed up computations.
-        return_loss : bool (default False), keyword only
+        return_loss : :obj:`bool` (default False), keyword only
```
NicolasGensollen, line 165: Could you double check that ? I thought you removed the loss from the returned values recently.

src/leaspy/algo/base.py
```diff
@@ -201,37 +205,57 @@ def load_parameters(self, parameters: dict):
 
         Parameters
         ----------
-        parameters : dict
+        parameters : :obj:`dict`
```
NicolasGensollen, line 208: The description above is weird: "The keys in the io which does not belong to the algorithm's parameters keys are ignored."

In the "input" maybe ??

src/leaspy/algo/base.py
```diff
@@ -276,16 +300,16 @@ def _display_progress_bar(
 
         Parameters
         ----------
-        iteration : int >= 0 or -1
+        iteration : :obj:`int` >= 0 or -1
```
NicolasGensollen, line 303: I think the description could be improved a bit:

"If a positive integer, it is the current iteration of the algorithm. If equals to '-1' then it is the final iteration."

Might also be worth checking in the code that this description is correct...

src/leaspy/algo/base.py
```diff
@@ -354,6 +378,17 @@ def __str__(self):
 
 
 def get_algorithm_type(name: Union[str, AlgorithmName]) -> AlgorithmType:
+    """
+    Return the algorithm type.
+
+    Parameters
+    ----------
+    name : :obj:`str`
```
NicolasGensollen, line 386:
```suggestion
    name : :obj:`str` or :class:`~leaspy.algo.base.AlgorithmName`
        The name of the algorithm.
```

src/leaspy/algo/base.py
```diff
@@ -363,6 +398,17 @@ def get_algorithm_type(name: Union[str, AlgorithmName]) -> AlgorithmType:
 
 
 def get_algorithm_class(name: Union[str, AlgorithmName]) -> Type[AbstractAlgo]:
+    """
+    Return the algorithm class.
+
+    Parameters
+    ----------
+    name : :obj:`str`
```
NicolasGensollen, line 406:
```suggestion
    name : :obj:`str` or :class:`~leaspy.algo.base.AlgorithmName`
         The name of the algorithm.
```

NicolasGensollen (commented):
Thanks @caglayantuna !

Looks good, I made a few small suggestions.

src/leaspy/algo/base.py
caglayantuna: Yes, I removed but not removed from the doc. Now, I removed

src/leaspy/algo/base.py
caglayantuna: Done and removed 2. "keys" from the sentence. Also I changed "wanted" to "requested" 2 lines below

src/leaspy/algo/base.py
caglayantuna: I don't think that current iteration becomes -1. i changed the lines and removed -1. WDYT?

src/leaspy/algo/base.py
caglayantuna: tests are failing because of this change but I couldn't understand where `return_loss` argument coming from

NicolasGensollen: There are a few places where this seems to be used:

https://github.com/aramis-lab/leaspy/blob/8132bcf57094a8b2ac1b9bfab6045e20ae86ad06/leaspy/api.py#L152

https://github.com/aramis-lab/leaspy/blob/8132bcf57094a8b2ac1b9bfab6045e20ae86ad06/tests/functional_tests/api/test_api.py#L60-L64

https://github.com/aramis-lab/leaspy/blob/8132bcf57094a8b2ac1b9bfab6045e20ae86ad06/tests/unit_tests/algo/others/test_constant_prediction_algo.py#L89

https://github.com/aramis-lab/leaspy/blob/8132bcf57094a8b2ac1b9bfab6045e20ae86ad06/tests/functional_tests/api/test_api_personalize.py#L39

I suggest you revert this change from this PR such that we can finish it while staying in scope (documentation of algo base). We'll open a dedicated issue and PR for this.

src/leaspy/algo/base.py
```diff
@@ -276,16 +292,16 @@ def _display_progress_bar(
 
         Parameters
         ----------
-        iteration : int >= 0 or -1
+        iteration : :obj:`int` >= 0
```
NicolasGensollen, line 295: The second line of the function is: 

```python
if iteration == -1:
```

So it seems that this can be "-1" and that something special is done is that case.

src/leaspy/algo/base.py
caglayantuna: thanks for your comments. I put -1 again but I din't change the description as it is not the final iteration. What do you think about;

""If a positive integer, it is the current iteration of the algorithm. If equals to '-1' then it initialises the bar.""

NicolasGensollen: Yes, I agree it's not very intuitive since -1 usually refers to last element in Python.
Your suggestion seems good to me.

caglayantuna: I added this new sentence. I think this PR is ready to merge

PR #331, opened 2025-03-07
remove return loss

caglayantuna:
This PR removes remaining `return_loss` from the functions and docs.


NicolasGensollen (approved):
LGTM, thanks @caglayantuna !

PR #346, opened 2025-04-01
Sk/mixture v2

KaisaridiSofia:
The algorithm works.

There are still some problems with the results. It ends "trapped" in the initial values for the probabilities or in placing all individuals in one cluster. I am working on fixing it. But from now on I think it's a computational/mathematical problem.


caglayantuna:
I worked on this PR to solve the remaining issues. Currently, fit and personalize functions are working if we ignore FitOutPut manager. In FOM, we use [compute_prior_trajectory](https://github.com/aramis-lab/leaspy/blob/v2/src/leaspy/models/abstract_model.py#L603) function to calculate mean. This function returns an error because of 2 reasons;

1. Data type of individual parameters (probs, tau, xi, sources) in mixture model is `float64` but it should be `float32`.
2. Size of these parameters are not the same after using [put_individual_latent_variables](https://github.com/aramis-lab/leaspy/blob/v2/src/leaspy/variables/state.py#L425). For example; size of tau in logistic : torch.Size([1, 1]) and in mixture : torch.Size([1]). This could be a problem as well.

PR #347, opened 2025-04-01
Add docstrings to simulate

maylistran01:
## TEST
Start of the docstrings for simulate algo.

### DIFF
--- a/src/leaspy/algo/simulate/simulate.py
+++ b/src/leaspy/algo/simulate/simulate.py
@@ -148,12 +148,12 @@ class SimulationAlgorithm(AbstractAlgo):
                 raise LeaspyAlgoInputError("Dataframe has null value in column TIME")
 
     ## --- SET PARAMETERS ---
-    def save_parameters(self, model, path_save):  # TODO
-        total_params = {"study": self.param_study, "model": model.parameters}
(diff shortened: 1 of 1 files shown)


src/leaspy/algo/simulate/simulate.py
```diff
@@ -204,6 +204,21 @@ def run_impl(self, model):
 
     ## ---- IP ---
     def get_ip_rm(self, model):
+
+         """
```
NicolasGensollen, line 208: The indentation is wrong, that's why you get an error during the documentation build in the CI.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -123,31 +128,90 @@ def _check_params(self, requirements):
...
+        """Set parameters related to the study based on visit type.
+
+        This function initializes the `param_study` attribute with relevant
+        parameters depending on the visit type of the object. It handles
+        three different visit types: 'dataframe', 'regular', and 'random',
+        each requiring a different set of input parameters.
+
+        Parameters
+        ----------
+        dict_param : :obj:`dict`
```
maylistran01, line 166: This params need to be specified in AlgorithmSettings. Should I add docstring in settings.py (list is not exhaustive there though)?

PR #350, opened 2025-04-07
Clean docstring of utilities

lea-aguilhon:
This PR updates docstrings of models/utilities files as it is mentioned in issue https://github.com/aramis-lab/leaspy/issues/122.

### DIFF
--- a/src/leaspy/models/utilities.py
+++ b/src/leaspy/models/utilities.py
@@ -22,6 +22,16 @@ __all__ = [
 def tensor_to_list(x: Union[list, torch.Tensor]) -> list:
     """
     Convert input tensor to list.
+
+    Parameters
+    ----------
+    x : :obj:`(Union[list, torch.Tensor])`
+        Input tensor or list.
+
+    Returns
+    -------
+    :obj:`list` :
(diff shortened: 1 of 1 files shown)


src/leaspy/models/utilities.py
```diff
@@ -213,6 +263,23 @@ def get_log_velocities(
 
 
 def torch_round(t: torch.FloatTensor, *, tol: float = 1 << 16) -> torch.FloatTensor:
-    """Round provided tensor."""
+    """
+    Round values to ~ 10**-4.8
```
caglayantuna, line 267: do we round all the values to this value? I think that we need a better explanation for this function. Could you check if you have time what are we doing with this function exactly?

caglayantuna:
thanks @lea-aguilhon , I have only one comment, other changes are ok to me

src/leaspy/models/utilities.py
lea-aguilhon: I checked it and change it in my new commits.

lea-aguilhon:
@caglayantuna I don't understand why the readthedocs check failed.

PR #353, opened 2025-04-09
Update Docstring data

lea-aguilhon:
This PR updates docstrings of data file files as it is mentioned in issue https://github.com/aramis-lab/leaspy/issues/122.


caglayantuna:
Thanks @lea-aguilhon! It seems good but in [rendered version](https://leaspy--353.org.readthedocs.build/en/353/reference/api/leaspy/io/data/data/index.html), we have some problems for List and Dict. I couldn't find the solution for now but I am writing here to find later.

lea-aguilhon:
I've resoled the issue with List and Doc, normally the rendered version is good now.

caglayantuna:
It seems better but we still don't have a link to `dict` and `list` [here](https://docs.python.org/3/library/functions.html). For `int`, it is working as expected. We can check and if it is not possible we can keep like this.

lea-aguilhon:
I think it is because Dict and List here are object define in leaspy and not dict and list python object

PR #357, opened 2025-04-17
Simulate final v2

lea-aguilhon:
### Creation simulate.py :

**Generate visit times**

Visit times are computed according to the type of visit provided: “random”, “regular”, and “dataframe”. 
 
For “regular” visit type:  
Visits are computed with regular spacing, given by regular_visit that is the time delta between two visits. 
- First visit is sampled from a normal distribution N(fv_mean, fv_std) 
- Total follow-up time is sampled form a normal distribution N(tf_mean,tf_std) 
 
For “dataframe” visit type: 
Visits are given for each ID by the user in a dataframe and this will be use for the simulation. 
 
For “random” visit type: 
- First visit is sampled from a normal distribution N(fv_mean, fv_std) 
-  Total follow-up time is sampled form a normal distribution N(tf_mean,tf_std) 
- Time delta between two visits is sampled from a normal distribution N(distv_mean, distv_std) 
 

**Generate individual parameters from the model parameters:** 

Based on the provided leaspy trained model and its parameters (in _italic)_ ta, we sample individual parameters for each ID:  

Tau is sampled from the normal distribution N_(tau_mean, tau_std)_  
Xi is sampled from the normal distribution N(_xi_mean, xi_std)_ 
Each source is sampled from the normal distribution N(0,1) and standardized after. The number of sources is given by the number of sources of the model.  
Space shifts are computed with the given _mixing matrix A_ and the computed sources s_i: w_i= A*s_i 
This is done in the function get_ip_rm.  

**Estimate observations with the new parameters:**  

Each ID has his new individual parameters. Estimate function of Leaspy is used to compute the observations based on these individual parameters (xi, tau, and the sources).  

First, the function get_leaspy_model loads the Leaspy model to be able to apply .estimate function. The estimate function compute the observations, which we clip between 0 and 1. These are the observations for each ID at each TIME, without noise.  

Then, we add a beta noise to these observations, using the mean and variance parametrization for alpha and beta. Some of the observations (very few) have an alpha/beta <0 which is not possible. It happens when the observation is super close to 1 (mu = 0.998). In that case, we can’t add a beta noise. -> still needs to be discussed.  

This is done in the function generate_dataset. 

### Modify the API to correspond simulate 

**1)** default_simulation.json was changed so that the input parameters correspond to the new simulate (visit_parameters, features) 
API was changed for an easier approach by the users:  

**Previously:**  

**2)** simulation_settings = AlgorithmSettings('simulation', seed=0, number_of_subjects=200, mean_number_of_visits=10, std_number_of_visits=0) 
simulated_data = leaspy_logistic.simulate(individual_parameters, data, simulation_settings) 

 
**Now:**  
simulation_settings = AlgorithmSettings( "simulation", seed=0, features=["MDS1_total", "MDS2_total", "MDS3_off_total", 'SCOPA_total','MOCA_total','REM_total','PUTAMEN_R','PUTAMEN_L','CAUDATE_R','CAUDATE_L'], visit_parameters= visits_params )  
simulated_data = leaspy_logistic.simulate(simulation_settings) 

No need to put the individual_parameters and the data in simulate function -> they are already in leaspy_logistic.  
Visit parameters contains all the information linked to the number/spacing of visits.  

## Where should the reviewer start ? 
Simulate.py

## How can the code be tested ? 
`from leaspy.algo import AlgorithmSettings`
`from leaspy.api import Leaspy `
`from leaspy.datasets import load_dataset `
`from leaspy.io.data import Data `
`parkinson_df = load_dataset("parkinson-multivariate") `
`data = Data.from_dataframe(parkinson_df) `
`leaspy_logistic = Leaspy('logistic', dimension = 10, source_dimension=2) `
`fit_settings = AlgorithmSettings('mcmc_saem', seed=0, n_iter=1000) `
`leaspy_logistic.fit(data, fit_settings) `
`visits_params = {'pat_nb':200, `
`'visit_type': "regular", `
`'regular_visit': 1, `
`'fv_mean' : 0.,  `
`'fv_std' : 0.4,  `
`'tf_mean' : 11,  `
`'tf_std' : 0.5,  `
`'distv_mean' : 2/12,  `
`'distv_std' : 0.75/12, `
`} `
`simulation_settings = AlgorithmSettings( `
`"simulation", `
`seed=0, `
`features=["MDS1_total", "MDS2_total", ``"MDS3_off_total",'SCOPA_total','MOCA_total','REM_total','PUTAMEN_R','PUTAMEN_L','CAUDATE_R','CAUDATE_L'], `
`visit_parameters= visits_params `
`) `
`simulated_data = leaspy_logistic.simulate(simulation_settings) `
`print(simulated_data.data.to_dataframe().set_index(['ID', 'TIME']).head()) `

## When is the MR due for? (review deadline) 
Mid may 

## What issues are linked to the MR ? 
None 

## What’s next?  
Unit tests improvement  
Add events?

### DIFF
--- a/docs/algorithms.md
+++ b/docs/algorithms.md
@@ -113,29 +113,34 @@ To run a simulation, the following variables are required:
 >>> from leaspy.algo import AlgorithmSettings
 >>> visits_params = {
         'patient_nb': 200,
-        'visit_type': "regular",
-        'regular_visit': 1,
+        'visit_type': "random",
         'first_visit_mean': 0.,
         'first_visit_std': 0.4,
         'time_follow_up_mean': 11,
         'time_follow_up_std': 0.5,
         'distance_visit_mean': 2 / 12,
         'distance_visit_std': 0.75 / 12,
+        'distance_visit_min': 1/365
     }
 >>> simulated_data = model.simulate( 
          algorithm="simulate", 
          features=["MDS1_total", "MDS2_total", "MDS3_off_total"],
          visit_parameters= visits_params
     )
-Simulate with `simulation` took: 3s
 >>> print(simulated_data.data.to_dataframe().set_index(['ID', 'TIME']).head())
-              MDS1_total   MDS2_total   MDS3_total
-ID     TIME
- 0   48.092     0.113991     0.344018     0.228870
-     49.092     0.177138     0.291440     0.332267
-     50.092     0.256084     0.253093     0.314241
-     51.092     0.174059     0.344921     0.313776
-     52.092     0.245289     0.400363     0.310065
+ ID  TIME  MDS1_total  MDS2_total  MDS3_off_total  SCOPA_total  MOCA_total  \
+  0  63.0    0.130888    0.220548        0.186086     0.083651    0.088756   
+     64.0    0.138080    0.039211        0.289588     0.034846    0.047147   
+     65.0    0.228149    0.068744        0.151979     0.141604    0.131976   
+     66.0    0.208679    0.112899        0.202224     0.192716    0.067183   
+     67.0    0.290484    0.252141        0.255622     0.240425    0.115898   
+
+   REM_total  PUTAMEN_R  PUTAMEN_L  CAUDATE_R  CAUDATE_L  
+   0.555283   0.808789   0.685063   0.546174   0.467885  
+   0.660931   0.758014   0.640209   0.541839   0.474202  
+   0.766028   0.941519   0.738120   0.643509   0.549832  
+   0.671021   0.796510   0.930209   0.657473   0.622322  
+   0.791594   0.955246   0.844813   0.677306   0.638281  
 ```
 
 ### Output


src/leaspy/algo/data/default_simulation.json
```diff
@@ -4,19 +4,9 @@
...
-    "min_number_of_visits": 1,
-    "max_number_of_visits": null,
-    "delay_btw_visits": 0.5,
     "cofactor": null,
     "cofactor_state": null,
-    "sources_method": "full_kde",
-    "reparametrized_age_bounds": null,
-    "features_bounds": false,
-    "features_bounds_nb_subjects_factor": 10
+    "features": null,
```
caglayantuna, line 9: is this "null" normal? when do we use this json file?

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+                f"Visit type need to be a string and not : {type(self.visit_type).__name__}"
+            )
+        try:
+            VisitType(self.visit_type)
+        except ValueError as e:
+            allowed_types = [vt.value for vt in VisitType]
+            raise LeaspyAlgoInputError(
+                f"Invalid visit type : '{self.visit_type}'. "
+                f"Authorized typz : {', '.join(allowed_types)}"
+            ) from e
```
caglayantuna, line 80: is it something you forgot here maybe?

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+    }
+
+    def __init__(self, settings):
+        super().__init__(settings)
+        self.features = settings.parameters["features"]
+        self.visit_type = settings.parameters["visit_parameters"]["visit_type"]
+        self._set_param_study(settings.parameters["visit_parameters"])
+        self._validate_algo_parameters()
+
+    ## --- CHECKS ---
```
caglayantuna, line 58: I would remove this comment from here. function name is already clear

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+        LeaspyAlgoInputError
+            If the features are not a list or if any of the features is not a string.
+        """
+
+        if not isinstance(self.features, list):
+            raise LeaspyAlgoInputError(
+                f"Features need to a be a list and not : {type(self.features).__name__}"
+            )
+        if len(self.features) == 0:
+            raise LeaspyAlgoInputError("List can't be empty")
```
caglayantuna, line 98: "Features can't be empty" could be better

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+                )
+
+            if df["TIME"].isnull().any():
+                raise LeaspyAlgoInputError("Dataframe has null value in column TIME")
+
+    ## --- SET PARAMETERS ---
+    # def _save_parameters(self, model, path_save):  # TODO
+    #     total_params = {"study": self.param_study, "model": model.parameters}
+    #     with open(f"{path_save}params_simulated.json", "w") as outfile:
+    #         json.dump(total_params, outfile)
```
caglayantuna, line 199: is it necessary to keep this commented function?

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+                "pat_nb": dict_param["pat_nb"],
+                "fv_mean": dict_param["fv_mean"],
+                "fv_std": dict_param["fv_std"],
+                "tf_mean": dict_param["tf_mean"],
+                "tf_std": dict_param["tf_std"],
+                "distv_mean": dict_param["distv_mean"],
+                "distv_std": dict_param["distv_std"],
+            }
+
+    ## ---- SIMULATE ---
```
caglayantuna, line 289: we can remove this comment as well

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+        -------
+        result_obj : :class:`Result`
+            An object containing:
+            - `data`: Simulated longitudinal dataset (`Data` object),
+            - `individual_parameters`: The individual parameters used in simulation,
+            - `noise_std`: Noise standard deviation used in the simulation.
+        """
+
+        # if seed is not None:
+        #     np.random.seed(seed)
```
caglayantuna, line 318: if we won't use, lets remove it. if you think that, it will be updated later, then we can keep it.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+            }
+
+        elif self.visit_type == "regular":
+            self.param_study = {
+                "pat_nb": dict_param["pat_nb"],
+                "regular_visit": dict_param["regular_visit"],
+                "fv_mean": dict_param["fv_mean"],
+                "fv_std": dict_param["fv_std"],
+                "tf_mean": dict_param["tf_mean"],
+                "tf_std": dict_param["tf_std"],
```
caglayantuna, line 275: Having `patient_number`, `first_visit_mean`, etc, is not so long. WDYT? If it causes too many changes in your PR, we can keep like this.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+        # Convert the result to a DataFrame
+        df_wn = pd.DataFrame(
+            result.T,
+            columns=[f"w_{i}" for i in range(len(self.features))],
+            index=df_ip_rm.index,
+        )
+
+        return pd.concat([df_ip_rm, df_wn], axis=1)
+
+    # ---- MODEL ---
```
caglayantuna, line 424: we can remove this comment as well

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+        Returns
+        -------
+        None
+            This method updates the `leaspy` attribute in-place.
+        """
+
+        self.leaspy = Leaspy("logistic", source_dimension=model.source_dimension)
+        self.leaspy.model = model
+
+    ## ---- RM ---
```
caglayantuna, line 446: same, we can remove it

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+    def _generate_dataset(
+        self, model, dict_timepoints: dict, df_ip_rm: pd.DataFrame
+    ) -> pd.DataFrame:
+        """
+        Generate a simulated dataset based on simulated individual parameters and model timepoints.
+
+        This method simulates observations using estimate function of the Leaspy model. The latter estimates
+        values based on the simulated individual parameters: xi, tau and the sources.
+        It then adds a beta noise to the simulated values.
+        Visits too close are dropped.
```
caglayantuna, line 533: "Visits too close to each other are removed" . It is not true actually, one of them should be removed. Right? But which one?

Maybe we can give more information here. How close? I see round(3) in the function. How did we choose it?

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+        df_ip_rm : pd.DataFrame
+            DataFrame containing the simulated individual parameters (e.g., 'xi', 'tau', and sources)
+            for each individual, used in generating the simulated data.
+
+        Returns
+        -------
+        pd.DataFrame
+            A DataFrame containing the simulated dataset with ["ID","TIME] as the index
+            and features as columns. The dataset includes both the generated values,
+            with visits that are too close to each other dropped.
```
caglayantuna, line 553: I prefer to use `remove` instead of `drop`. I understand it is coming from pandas but we use `remove` for better explanation.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,625 @@
...
+            if model.parameters["noise_std"].numel() == 1:
+                mu = df_long[feat + "_no_noise"]
+                var = model.parameters["noise_std"].numpy() ** 2
+            else:
+                mu = df_long[feat + "_no_noise"]
+                var = model.parameters["noise_std"][i].numpy() ** 2
+
+            # Mean and sample size (P-E simulations)
+            # alpha_param = mu * var
+            # beta_param = (1 - mu) * var
```
caglayantuna, line 588: if these commented lines are not useful for the future, we can remove.

tests/unit_tests/algo/simulate/test_simulate.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Check times are increasing with variability
+        for id in df_sim["ID"].unique():
+            times = df_sim.loc[df_sim["ID"] == id, "TIME"].values
+            diffs = np.diff(times)
+            self.assertTrue((diffs > 0).all())
+            self.assertGreater(np.std(diffs), 0)
+
+    def test_dataframe_visits(self):
+        model_loaded = self.model_loaded
```
caglayantuna, line 135: we can remove this and use `self.model.loaded` in line 157.

caglayantuna (commented):
thanks @maylistran01 and @lea-aguilhon for this PR. It seems very nice. Unit tests are already good. 

I have some comments and questions. After your answers, we can merge it

JulietteOrtholand:
Thanks @lea-aguilhon and @maylistran01 !  I have not read the code yet but I have a few questions: 

- Why do you need to clip between 0 and 1? Don’t you simulate with a logistic?
- by "No need to put the individual_parameters and the data in simulate function" you mean the parameters of the model? aren’t individual parameters sampled?

maylistran01:
Thank you @JulietteOrtholand for the comment.

- You're right, this is not needed as we are computing simulated observations with .estimate function which gives values between 0 and 1. 
- I wasn't clear in this sentence: 
In the actual Leaspy simulate function, here's how we simulate data:
`simulated_data = leaspy_logistic.simulate(**individual_parameters, data,** simulation_settings). The users had to load the data, train the model and then load the `individual_parameters` and create a dedicated variable in his code before simulating data:
`
individual_parameters = Loader.load_individual_parameters('parkinson-putamen-train')
data = Data.from_dataframe(putamen_df.xs('train', level='SPLIT'))
`
We changed it to look like the api, and so that the user just need to train the model and the simulate the data directly : 
`simulated_data = **leaspy_logistic.simulate**(simulation_settings)` -> That's what I meant by "No need to put the individual_parameters and the data in simulate function". But of course the individual parameters are sampled for each patient from the model parameters.

Let me know if our answer is clear enough !

src/leaspy/algo/data/default_simulation.json
maylistran01: We use this json file in [settings.py](https://github.com/aramis-lab/leaspy/blob/v2/src/leaspy/algo/settings.py), row 328.
I set to "null" because there is no pre-requisites for features (it depends on the data we are using, could be ["ALSFRS-r", "BMI"] for instance.

src/leaspy/algo/simulate/simulate.py
maylistran01: Yes, I still need to update it so that we can save the parameters of simulation.

src/leaspy/algo/simulate/simulate.py
maylistran01: **More information in the docstring**
We'll modify the docstring as follow : "Visits too close are droped." -> "If the visits time are too close to each other, we keep only the first occurence. By "too close", we mean if the visits happened on the same day."

**How did we choose it?**
We consider that 2 visits didn't happen the same day. As TIME is in years: 
TIME.round(3) : 0.001 years ≈ 0.36525 days
TIME.round(2) : 0.01 years ≈ 3.6525 days
TIME.round(1) : 0.1 years ≈ 36.525 days
-> to make sure there is no more than 1 visit in a day, we need round(3).

src/leaspy/algo/simulate/simulate.py
maylistran01: They might be useful in the future if we want to change the beta parametrization. Can we keep it?

maylistran01 (commented):
Hi @caglayantuna, 
Thanks for the review ! 
Here are our answers to the comments we still want to discuss. We took into account all the other comments, and will adapt our code accordingly.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        -------
+        result_obj : :class:`Result`
+            An object containing:
+            - `data`: Simulated longitudinal dataset (`Data` object),
+            - `individual_parameters`: The individual parameters used in simulation,
+            - `noise_std`: Noise standard deviation used in the simulation.
+        """
+
+        # Simulate RE for RM
+        df_ip_rm = self._get_ip_rm(model)
```
NicolasGensollen, line 316: A general advise is to avoid naming tricks like these because it puts unnecessary cognitive load onto the reader. When I read this, either I don't understand anything, or I remember that you defined these acronyms somewhere, and I must either remember them or find this definition. And I will need to do that every time I encounter these variables.

Modern IDEs offer code completion such that defining long variable names isn't a huge burden for the developer any longer.

Reading something like this:

```python
individual_parameters_from_repeated_measurements = (
    self._get_individual_parameters_from_repeated_measurements(model)
)
```

will be much much easier to understand even though it is longer.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+
+        # Get Leaspy model
```
NicolasGensollen, line 318: You don't need comments, the code is clear enough.

```suggestion
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        self._get_leaspy_model(model)
+
+        # Generate visits ages
```
NicolasGensollen, line 321:
```suggestion
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        dict_timepoints = self._generate_visit_ages(df_ip_rm)
+
+        # Get all visits observations
```
NicolasGensollen, line 324:
```suggestion
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+                "patient_number": dict_param["patient_number"],
+                "first_visit_mean": dict_param["first_visit_mean"],
+                "first_visit_std": dict_param["first_visit_std"],
+                "time_follow_up_mean": dict_param["time_follow_up_mean"],
+                "time_follow_up_std": dict_param["time_follow_up_std"],
+                "distance_visit_mean": dict_param["distance_visit_mean"],
+                "distance_visit_std": dict_param["distance_visit_std"],
+            }
+
+    def run_impl(self, model) -> Result:
```
NicolasGensollen, line 288: The type hint is incomplete here.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+
+        tau_rm = torch.tensor(
+            np.random.normal(
+                model.parameters["tau_mean"],
+                model.parameters["tau_std"],
+                self.param_study["patient_number"],
+            )
+        )
+
+        if self.visit_type == "dataframe":
```
NicolasGensollen, line 375: You can simplify this a bit:

```python
if self.visit_type == VisitType.DATAFRAME:
    columns = [str(i) for i in self.param_study["df_visits"]["ID"].unique()]
else:
    columns = [str(i) for i in range(0, self.param_study["patient_number"])]
individual_parameters_from_repeated_measurements = pd.DataFrame(
    [xi_rm, tau_rm],
    index=["xi", "tau"],
    columns=columns,
).T
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        for i in range(model.source_dimension):
+            df_ip_rm[f"sources_{i}"] = torch.tensor(
+                np.random.normal(0.0, 1.0, self.param_study["patient_number"]),
+                dtype=torch.float32,
+            )
+            df_ip_rm[f"sources_{i}"] = (
+                df_ip_rm[f"sources_{i}"] - df_ip_rm[f"sources_{i}"].mean()
+            ) / df_ip_rm[f"sources_{i}"].std()
+
+        pat = torch.stack(
```
NicolasGensollen, line 399: What is "pat" ?

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+            [
+                torch.tensor(df_ip_rm[f"sources_{i}"].values, dtype=torch.float32)
+                for i in range(model.source_dimension)
+            ],
+            dim=1,
+        )
+        mat = model.state.get_tensor_value("mixing_matrix")
```
NicolasGensollen, line 406:
```suggestion
        mixing_matrix = model.state.get_tensor_value("mixing_matrix")
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        result = torch.matmul(mat.transpose(0, 1), pat.transpose(0, 1))
+
+        # Convert the result to a DataFrame
```
NicolasGensollen, line 409:
```suggestion
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        df_wn = pd.DataFrame(
```
NicolasGensollen, line 410: Isn't there a more explicit name for this variable ?

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        model : :class:`Leaspy`
+            A pre-trained Leaspy model to be used for simulation (compute observations).
+
+        Returns
+        -------
+        None
+            This method updates the `leaspy` attribute in-place.
+        """
+
+        self.leaspy = Leaspy("logistic", source_dimension=model.source_dimension)
```
NicolasGensollen, line 436: I think we should avoid using the `Leaspy` class. @caglayantuna and I were actually planning to delete it because it is very confusing to have a giant class with the same name as the package doing everything. Instead, the package should define a public API with proper modules. 
So, if you need to define a logistic model here, I would expect something like this:

```python
from leaspy.models import LogisticMultivariateModel

self.model = LogisticMultivariateModel(source_dimension=model.source_dimension)
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+        dict
+            Dictionary mapping individual IDs to a list of visit ages (floats).
+            - For 'dataframe': uses existing "TIME" values from `df_visits`.
+            - For 'regular': generates visits at fixed intervals.
+            - For 'random': generates visits with normally-distributed intervals.
+        """
+
+        df_ind = df.copy()
+
+        if self.visit_type == "dataframe":
```
NicolasGensollen, line 465: You can simplify this by returning early:

```python
if self.visit_type == VisitType.DATAFRAME:
    return self.param_study["df_visits"].groupby("ID")["TIME"].apply(list).to_dict()
# Then you can unindent everything
df_ind["AGE_AT_BASELINE"] = ....
```

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+            dict_timepoints = {}
+
+            for id_ in df_ind.index.values:
+                # Get the number of visit per patient
+                time = df_ind.loc[id_, "AGE_AT_BASELINE"]
+                age_visits = [time]
+
+                while time < df_ind.loc[id_, "AGE_FOLLOW_UP"]:
+                    if self.visit_type == "regular":
+                        time += self.param_study["regular_visit"]
```
NicolasGensollen, line 502: Is there a check somewhere to avoid having an infinite loop here ?

For example, if `self.param_study["regular_visit"] == 0` or if `self.param_study["distance_visit_mean"] == 0` and `self.param_study["distance_visit_std"] == 0`.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+                    np.random.normal(
+                        self.param_study["first_visit_mean"],
+                        self.param_study["first_visit_std"],
+                        self.param_study["patient_number"],
+                    ),
+                    index=df_ind.index,
+                )[0]
+            )
+
+            df_ind["AGE_FOLLOW_UP"] = df_ind["AGE_AT_BASELINE"] + np.random.normal(
```
NicolasGensollen, line 486: Sorry if this is a stupid question, but the random variable that you draw can be negative, right ?

In this case, you'd have `AGE_FOLLOW_UP < AGE_AT_BASELINE`, would that make sense ?

NicolasGensollen (commented):
Thanks @lea-aguilhon for the work !

I made a first pass with some comments and suggestions along the way.

src/leaspy/algo/simulate/simulate.py
maylistran01: Right, I will change to: 
`   def run_impl(self, model: AbstractModel) -> Result:
`
One question: I also need to change the docstring from :
model : :class:`Leaspy` 
to 
model : :class:`~.models.abstract_model.AbstractModel`, right?

maylistran01 (commented):
Thank you @NicolasGensollen for this review ! 
Here are my answers regarding the comments on the parts I worked on. Lea will answer to the others.

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: Thanks for the comment, I've changed it to the absolute value to avoid this problem.

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: I'll add these checks

src/leaspy/algo/simulate/simulate.py
caglayantuna: ok for me

src/leaspy/algo/simulate/simulate.py
NicolasGensollen: Maybe change the name of the variable for something a little less obscure like "patient_source_values" or "patient_source_values_matrix", WDYT ?

src/leaspy/algo/simulate/simulate.py
NicolasGensollen: It might be clear for someone who did read the related scientific papers and who payed attention to equations. For most readers however, this is not obvious at all. Calling this variable "space_shifts" avoid having to do this mental mapping between concepts and equations variable names.

src/leaspy/algo/simulate/simulate.py
NicolasGensollen: Thanks for the explanations @maylistran01 

Ideally, the `estimate` method would be a Model method or an independent function accepting a model and other parameters but there is some work to do to enable this. Moreover, it's not the point of this PR, so I believe we can leave it like that for now and @caglayantuna or I will work on removing the Leaspy class later.

JulietteOrtholand (commented):
Thank you very much @lea-aguilhon and @maylistran01 for this PR! I made a few comments, happy to discuss that further if needed :)

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,631 @@
...
+            Dictionary containing parameters required for the study. The
+            expected keys vary depending on the visit type:
+
+            - If `visit_type` is "dataframe":
+                - 'df_visits' : :obj:`pandas.DataFrame`
+                    DataFrame of visits, with a column "ID" and a column 'TIME'.
+                TIME and number of visits for each simulated patients (with specified ID)
+                are given by a dataframe in dict_param.
+
+            - If `visit_type` is "regular":
```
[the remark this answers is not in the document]
lea-aguilhon: We separate it to be more easy for a user, but maybe we need to rename those type of parameters to be more explicit ? 
random > random interval 
regular > fixed interval

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,631 @@
...
+
+        simulated_data = Data.from_dataframe(df_sim)
+        result_obj = Result(
+            data=simulated_data,
+            individual_parameters=individual_parameters_from_repeated_measurements,
+            noise_std=model.parameters["noise_std"].numpy() * 100,
+        )
+        return result_obj
+
+    def _get_individual_parameters_from_repeated_measurement(
```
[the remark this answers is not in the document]
lea-aguilhon: Ok we'll change it

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: I've removed it

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: Yes, thanks, it has been removed

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: It has been changed

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: We changed it for more explicit version

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: It has been done

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: For now we don't use it, so we removed it

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,618 @@
...
+                    Standard deviation of distance_visits: std time delta between two visits.
+                Time delta between 2 visits is drawn in a normal distribution N(distance_visit_mean, distance_visit_std).
+
+        Returns
+        -------
+        None
+            This method updates the `param_study` attribute of the instance in-place.
+        """
+
+        if self.visit_type == "dataframe":
```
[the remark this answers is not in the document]
lea-aguilhon: We changed it to use enumeration

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: It has been done

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: We changed it for more explicit version

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: It has been removed

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: It has been removed

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: It has been removed

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: We changed it accordingly

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: We changed it for patient_source_values_matrix

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: we changed it

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: We removed it

src/leaspy/algo/simulate/simulate.py
lea-aguilhon: We changed it for space_shifts

src/leaspy/algo/simulate/simulate.py
maylistran01: Thanks @NicolasGensollen for the explanations. Would you prefer to close the thread or open an issue for that? Feel free to open the issue if you have more details in mind you'd like to include.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,631 @@
...
+        model : :class:~.models.abstract_model.AbstractModel
+            A pre-trained Leaspy model to be used for simulation (compute observations).
+
+        Returns
+        -------
+        None
+            This method updates the `leaspy` attribute in-place.
+        """
+
+        self.leaspy = Leaspy("logistic", source_dimension=model.source_dimension)
```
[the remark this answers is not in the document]
maylistran01: Linked to this thread: https://github.com/aramis-lab/leaspy/pull/357/files#r2073054953
As discussed last week, we can close solve this comment.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,631 @@
...
+        individual_parameters_from_repeated_measurements: pd.DataFrame,
+    ) -> pd.DataFrame:
+        """
+        Generate a simulated dataset based on simulated individual parameters and model timepoints.
+
+        This method simulates observations using estimate function of the Leaspy model. The latter estimates
+        values based on the simulated individual parameters: xi, tau and the sources.
+        It then adds a beta noise to the simulated values.
+        If the visits time are too close to each other, we keep only the first occurrence.
+        By "too close", we mean if the visits happened on the same day.
```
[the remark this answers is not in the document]
maylistran01: Ok, we modified the function so that in the dict visit_parameters, the users can add a value for "min_days_spacing_between_visits". By default, it is set to 1 -> 2 visits can't happen the same day.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,631 @@
...
+                mu = df_long[feat + "_no_noise"]
+                var = model.parameters["noise_std"][i].numpy() ** 2
+
+            # Mean and sample size (P-E simulations)
+            # alpha_param = mu * var
+            # beta_param = (1 - mu) * var
+
+            # Mean and variance parametrization
+            alpha_param = mu * ((mu * (1 - mu) / var) - 1)
+            beta_param = (1 - mu) * ((mu * (1 - mu) / var) - 1)
```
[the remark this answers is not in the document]
maylistran01: Ok, we modified the function. 
We keep the mean and variance parametrization, but we made sure that `var < mu(a-mu)`. 

When it is not the case (example: mu=0.999 and var = 0.002), we slightly reduce the variance with: 
`adj_var = np.minimum(var, 0.99 * mu * (1 - mu))`. Like this, the beta distribution is well parametrized and we make sure it works all the time.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,631 @@
...
+        each requiring a different set of input parameters.
+
+        Parameters
+        ----------
+        dict_param : :obj:`dict`
+            Dictionary containing parameters required for the study. The
+            expected keys vary depending on the visit type:
+
+            - If `visit_type` is "dataframe":
+                - 'df_visits' : :obj:`pandas.DataFrame`
```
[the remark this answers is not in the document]
maylistran01: As discussed in Leaspy meeting, we don't need this as it is too close to estimate.

maylistran01 (commented):
Hey @JulietteOrtholand ! 
Thanks for the review. I answered and modified the points we discussed last week in Leaspy meeting. @lea-aguilhon will push the updated code.

src/leaspy/algo/simulate/simulate.py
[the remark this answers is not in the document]
lea-aguilhon: We removed the regular method to use only random one.

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,631 @@
...
+            A Leaspy model instance containing parameters set after training,
+            namely the mean and standard deviation values for xi, tau, and the mixing matrix.
+
+        Returns
+        -------
+        pd.DataFrame
+            A DataFrame indexed by individual IDs, containing:
+            - simulated 'xi' and 'tau': Individual parameters sampled from model distributions.
+            - simulated 'sources_X': Latent source components.
+            - simulated 'w_X': space shifts derived from the mixing matrix and sources.
```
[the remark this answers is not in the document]
lea-aguilhon: We made the changes

src/leaspy/algo/simulate/simulate.py
NicolasGensollen: I opened #375 for this specific task.

Closing this thread....

JulietteOrtholand:
Thank you very much @lea-aguilhon and @maylistran01 for integrating my comments! I still left some threads opened, but there are only minor reviews! I don't know if I get the chance to approve this MR through git, but you have my approval to merge in the future (and close my remaining threads) :) Again thanks a lot for this work!

JulietteOrtholand (commented):
sorry I forgot to publish the comments...

src/leaspy/algo/settings.py
```diff
@@ -306,6 +306,7 @@ def __init__(self, name: str, **kwargs):
         )
         if default_algo_settings_path.is_file():
             self._load_default_values(default_algo_settings_path)
+            # Dans le fichier où se trouve AlgorithmName, ajoutez :
```
maylistran01, line 309: Maybe we want to remove this

src/leaspy/algo/simulate/base.py
```diff
@@ -0,0 +1,82 @@
...
+        Returns
+        -------
+        result_obj : :class:`Result`
+            An object containing:
+            - `data`: Simulated longitudinal dataset (`Data` object),
+            - `individual_parameters`: The individual parameters used in simulation,
+            - `noise_std`: Noise standard deviation used in the simulation.
+        """
+
+        # Simulate RE for RM
```
maylistran01, line 56: Write the full words

src/leaspy/algo/simulate/simulate.py
```diff
@@ -0,0 +1,642 @@
...
+        Returns
+        -------
+        result_obj : :class:`Result`
+            An object containing:
+            - `data`: Simulated longitudinal dataset (`Data` object),
+            - `individual_parameters`: The individual parameters used in simulation,
+            - `noise_std`: Noise standard deviation used in the simulation.
+        """
+
+        # Simulate RE for RM
```
maylistran01, line 309: Same here

src/leaspy/models/base.py
```diff
@@ -609,11 +610,94 @@ def compute_individual_trajectory(
...
+                 'visit_type': "random",
+                # 'visit_type': "dataframe",
+                # "df_visits": df_test
+                 'first_visit_mean' : 0., #OK1
+                 'first_visit_std' : 0.4, #OK2
+                 'time_follow_up_mean' : 11, #OK
+                 'time_follow_up_std' : 0.5, #OK
+                 'distance_visit_mean' : 2/12, #OK # 1.
+                 'distance_visit_std' : 0.75/12, #OK # 6
+                 'min_spacing_between_visits': 1
```
maylistran01, line 682: It is ` 'min_spacing_between_visits': 1/365` if time is in years

src/leaspy/models/base.py
```diff
@@ -609,11 +610,94 @@ def compute_individual_trajectory(
...
         self,
-        individual_parameters: IndividualParameters,
-        data: Optional[Union[pd.DataFrame, Data, Dataset]] = None,
+        algorithm: Optional[Union[str, AlgorithmName]] = None,
+        algorithm_settings: Optional[AlgorithmSettings] = None,
+        algorithm_settings_path: Optional[Union[str, Path]] = None,
         **kwargs,
     ):
-        raise NotImplementedError("Simulation not implemented.")
+        r"""
```
maylistran01, line 618: Remove r?

src/leaspy/models/base.py
```diff
@@ -609,11 +610,94 @@ def compute_individual_trajectory(
...
+                # "df_visits": df_test
+                 'first_visit_mean' : 0., #OK1
+                 'first_visit_std' : 0.4, #OK2
+                 'time_follow_up_mean' : 11, #OK
+                 'time_follow_up_std' : 0.5, #OK
+                 'distance_visit_mean' : 2/12, #OK # 1.
+                 'distance_visit_std' : 0.75/12, #OK # 6
+                 'min_spacing_between_visits': 1
+                }
+        >>> simulated_data = simulated_data = model.simulate( algorithm="simulate", features=["MDS1_total", "MDS2_total", "MDS3_off_total", 'SCOPA_total','MOCA_total','REM_total','PUTAMEN_R','PUTAMEN_L','CAUDATE_R','CAUDATE_L'],visit_parameters= visits_params  )
```
maylistran01, line 684:
```suggestion
        >>> simulated_data = model.simulate( 
                     algorithm="simulate", 
                     features=["MDS1_total", "MDS2_total", "MDS3_off_total", 'SCOPA_total', 'MOCA_total', 'REM_total', 'PUTAMEN_R', 'PUTAMEN_L', 'CAUDATE_R', 'CAUDATE_L'],
                     visit_parameters=visits_params 
                )
```

src/leaspy/models/base.py
```diff
@@ -609,11 +610,94 @@ def compute_individual_trajectory(
...
+        >>> from leaspy.io.data import Data
+        >>> from leaspy.datasets import load_dataset, load_leaspy_instance, load_individual_parameters
+        >>> putamen_df = load_dataset("parkinson-putamen-train_and_test")
+        >>> data = Data.from_dataframe(putamen_df.xs('train', level='SPLIT'))
+        >>> leaspy_logistic = load_leaspy_instance("parkinson-putamen-train")
+        >>> visits_params = {'patient_number':200,
+                 'visit_type': "random",
+                # 'visit_type': "dataframe",
+                # "df_visits": df_test
+                 'first_visit_mean' : 0., #OK1
```
maylistran01, line 676: Remove the #OK?

maylistran01: And others # as well

caglayantuna:
Thanks @lea-aguilhon and @maylistran01  for your very good work and all of your answers. It seems ready to merge !

PR #361, opened 2025-05-05
Add support for bibliography

NicolasGensollen:
This PR proposes to add support for citations using the Bibtex format.


docs/mathematics.md
```diff
@@ -4,6 +4,13 @@
...
 
 ## Riemanian framework
-TODO
\ No newline at end of file
+TODO
+
+## References
+
+```{bibliography}
+:filter: docname in docnames
```
NicolasGensollen, line 15: This line makes sure only the references cited in this file are rendered here. With this, we can maintain a single bib file for the whole Leaspy documentation and only display the relevant citations on a page by page basis.

docs/references.bib
```diff
@@ -0,0 +1,8 @@
+@article{einstein1905,
```
NicolasGensollen, line 1: This file might become very very long if the documentation needs to make a lot of references. It's not a big deal, but it is very important to keep the entries sorted alphabetically to avoid chaos.

NicolasGensollen:
@JulietteOrtholand I'm gonna merge and will let you replace Einstein with more relevant citations. 

Let me know if this doesn't work for you.

PR #358, opened 2025-04-18
Documentation

JulietteOrtholand:
This PR is for the three documentation tasks I have to do:

- [x] Riemmanian context
- [x] literature using leaspy
- [x] Personnalisation

Issue #107, opened 2024-03-28
Convergence monitoring and MCMC - SAEM use


NicolasGensollen:
In GitLab by @JulietteOrtholand on Mar 28, 2024, 15:32

# Introduction

## Context
When using leaspy, you have to choose the number of iterations to estimate the model and hopefully to reach convergence of the MCMC-SAEM algorithm used. Until now this was done visually by looking at the curve of the value of parameters at each iteration, but this is not a robust method.

__Problem:__ How to assess convergence in leaspy?

## Background
Currently, there are a lot of issues related more or less to that subject:
- #65: question the value of the parameters set by default in the MCMC-SAEM algorithm and emphasise the need to find references to support these choices,
- #62: question specifically the annealing part, with some references on it and the implementation of oscillating annealing already introduced,
- #68: describe two needs: one to have a proper visualisation of what happened during the fit (maybe with tensor-board) and another to monitor properly convergence during and at the end of fit.
- #63: formulate the need for a metric to evaluate convergence, with some references,
- #27: question the fact that maybe we could stop the fit before the number of iterations set by the user is reached if we can monitor the convergence during the fit.

So if we summarize all this it results in different points:
1. Monitoring the convergence at the end of the fit,
2. Monitoring dynamically the convergence during the fit to stop earlier,
3. Have a proper visualisation of iterations,
4. Document and validate choices made for hyperparameters of MCMC-SAEM,
5. Improve the algorithm of estimation,

__Objective:__ Here the objective is to address specifically point 1 but it is closely related to points 2 and 4.

# MCMC & convergence monitoring

## Reminder on Markov chain Monte Carlo (MCMC)

MCMC algorithms are a class of methods allowing to sample from a complex distribution. If you are not familiar with MCMC I recommend you this [video](https://www.youtube.com/watch?v=WG9JMAytp2o) (17 min in French) or maybe this one shorter [video](https://www.youtube.com/watch?v=OTO1DygELpY) (8 minutes in english). Then what is important is that the chain evolution only depends on the previous state: $p(x_{t+1}|x_{t})$. 

[Put a paper ref + more info if needed after % MCMC-SAEM]

## Convergence monitoring

[...]

PR #367, opened 2025-05-13
[DOC] Documentation model evaluation

maylistran01:
This PR is for the documentation tasks from Models Evaluation part Léa and I have to do:

- [ ]  Convergence diagnosis

- [x] Fit metrics

- [x] Prediction metrics


docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+
+### BIC
+BIC (Bayesian Information Criterion) is similar to the AIC metric, but it also integrates the number of patients. It penalizes both the number of features and the number of patients.\cite{schwarz1978estimating}
+
+$$
+\text{BIC} = \log(\text{nb\_patient}) \cdot \text{features} - 2 \cdot \log(\text{likelihood})
+$$
+
+**Package:** `statsmodels.api`
+
```
JulietteOrtholand, line 58: Finally, you did not talk about the fact that the model is a mixed effect model? You did not cite the paper of [lavielle](https://projecteuclid.org/journals/electronic-journal-of-statistics/volume-8/issue-1/A-note-on-BIC-in-mixed-effects-models/10.1214/14-EJS890.full) on that, maybe could be interesting?

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+This section describes the main metrics used to evaluate the quality of predictions from the Leaspy non-linear mixed-effects model.
+
+#### Mean Absolute Error (MAE)
+
+$$
+MAE = \frac{1}{n} \sum_{i=1}^{n} |y_i - \hat{y}_i|
+$$
+
+MAE is more robust to outliers and provides a straightforward sense of typical prediction error.
```
JulietteOrtholand, line 74: Could you add a small code? + Maybe a reference? Even if it's wikipedia

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+$$
+
+MSE penalizes larger errors more heavily than MAE, making it more sensitive to outliers. \cite{willmott2005advantages}\cite{chai2014root}
+
+**Package:** `sklearn.metrics`
+
+#### Residual Q-Q Plot
+A graphical tool to assess whether residuals follow a normal distribution, which is an assumption in many mixed-effects models.\cite{fox2015applied}
+
+**Package:** `statsmodels.api`
```
JulietteOrtholand, line 89: as before could you put the lik to the documenattion and a small code how to use

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+#### Coefficient of Determination (R²)
+R² indicates how well the model explains the variance in the observed data. Higher values (closer to 1) suggest better performance.
+
+$$
+R^2 = 1 - \frac{\sum_{i=1}^{n} (y_i - \hat{y}_i)^2}{\sum_{i=1}^{n} (y_i - \bar{y})^2}
+$$
+
+In mixed-effects models, multiple R² variants exist (e.g., marginal vs. conditional R²) to account for fixed and random effects. \cite{nakagawa2012method}
+
+**Package:** `sklearn.metrics`
```
JulietteOrtholand, line 100: code+ link

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+
+### Events
+
+For the joint model, it is interesting to have prediction metrics on survival as well:
+
+#### Integrated Brier Score (IBS)
+The Integrated Brier Score (IBS) is a robust metric used to evaluate the predictive accuracy of survival models. In survival analysis, IBS quantifies how well a model predicts the timing of events by integrating the Brier Score, a measure of prediction error for probabilistic outcomes, across all observed time points. It compares the model’s predicted survival probabilities against actual observed outcomes, penalizing discrepancies between predictions and reality. IBS accounts for censored data, ensuring reliability in real-world datasets where not all events are fully observed. IBS describes how closely predictions match reality. A low IBS indicates superior predictive performance, as it reflects smaller cumulative errors over time.\cite{graf1999assessment}
+
+**Package:** `scikit-survival`
```
JulietteOrtholand, line 109: code + link to the specific function documentation

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+
+#### Cumulative Dynamic AUC
+Cumulative AUC (or time-dependent AUC) evaluates the model’s ability to discriminate between subjects who experience an event by a specific time and those who do not, based on their predicted risk scores. It focuses on ranking accuracy, ensuring high-risk subjects receive higher predicted probabilities than low-risk ones at each time point. Cumulative AUC describes how well risks are ordered. A high cumulative AUC means a strong discriminatory ability.  \cite{uno2007evaluating} \cite{hung2010estimation} \cite{lambert2014summary}
+
+**Package:** `scikit-survival`
```
JulietteOrtholand, line 114: code + link to the specific function documentation

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+#### Residual Q-Q Plot
+A graphical tool to assess whether residuals follow a normal distribution, which is an assumption in many mixed-effects models.\cite{fox2015applied}
+
+**Package:** `statsmodels.api`
+
+#### Coefficient of Determination (R²)
+R² indicates how well the model explains the variance in the observed data. Higher values (closer to 1) suggest better performance.
+
+$$
+R^2 = 1 - \frac{\sum_{i=1}^{n} (y_i - \hat{y}_i)^2}{\sum_{i=1}^{n} (y_i - \bar{y})^2}
```
JulietteOrtholand, line 95: Maybe if the documentation of the fonction you call is well done, no need for the equations?

docs/to_go_further.md
```diff
@@ -6,4 +6,4 @@
 ## What kind of scientific question could be answered with leaspy? 
 ### Covariates 
 Some papers have been published adressing such kind of questions [REF].
-[Small summary and an exemple of plot]
\ No newline at end of file
+[Small summary and an exemple of plot]
```
JulietteOrtholand, line 9: Why did you change that? Might cause conflict when merged with my PR... Could you rather delete that change and add a comment on my documentation PR if needed?

JulietteOrtholand (commented):
Thank you very much for this PR! The references and scientific contend is there, but I think this documentation could be improve by making a more explicit link with the code and how to actually do it :) Please let me know if my comments are not clear, and thanks again for this great work !

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
-# Models Selection and Evaluation
+# Models Evaluation
 
 ## Convergence diagnosis
 
+*We need to talk with Sofia in June to complete this part.*
+
+*At first glance, we would talk about looking at the convergence graphs saved in the logs (all the plots in convergence.pdf). WDYT?*
```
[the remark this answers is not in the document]
maylistran01: Ok, I added it in the general presentation of the PR, in the TODO part and I added the link to the issue 107, thanks for pointing there !

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
 
+*We need to talk with Sofia in June to complete this part.*
+
+*At first glance, we would talk about looking at the convergence graphs saved in the logs (all the plots in convergence.pdf). WDYT?* 
+
+
 ## Likelihood based metrics
 
-## Reconstruction metrics
+*Sofia and Gabrielle*
```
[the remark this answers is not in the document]
maylistran01: Done.

docs/models_evaluation.md
maylistran01: I thought we said in the meeting yesterday that we should add all the equations?

maylistran01 (commented):
Thanks Juliette for the review ! 
I made the first easy-fixes. I'll come back to you when we've added the code with the application of Leaspy + computation of the metrics + the ref to the documentation of the package.

docs/models_evaluation.md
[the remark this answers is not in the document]
JulietteOrtholand: Maybe it is out of scope of this PR, WDYT?

docs/models_evaluation.md
JulietteOrtholand: Yes, you are right, it's just that when reviewing the PR I was wondering how hard it might be to maintain the equations right for a function that is implemented by another package and that might change with time... So in a perfect world I think having the equations in the doc would be nice, but in a realistic world, wouldn't it be better to have no equation but a good reference toward them rather than wrong equation because the function has been updated by the other library... WDYT?

docs/models_evaluation.md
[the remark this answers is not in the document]
maylistran01: You mean that we should merge before solving that in June, right? 
If so, yes I agree. I'll just keep it in the comment of the PR to keep it in mind for now. Or is there a better place to write this TODO to not forget, in github?

docs/models_evaluation.md
maylistran01: To what extent do the equations in a package actually change? I would assume—perhaps incorrectly—that formulas like MAE or MSE remain constant. Does this assumption not apply to all the metrics listed here?

docs/models_evaluation.md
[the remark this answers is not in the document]
JulietteOrtholand: I think this is the use of the issues. As documentation issue #358 contains everything, I would simply update it to precise that, WDYT?

docs/models_evaluation.md
JulietteOrtholand: [Here](https://python.arviz.org/en/stable/api/generated/arviz.rhat.html) you have a function to compute a statistic on an MCMC chain. What you can see in the references is that the notion has been "invented" by Gelman and Rubin, yet the equation used is not the one of this version because methodological work has been done since then and a more robust version has been developed. So the package does not change the API but the black box behind to update it. You can also think about it with the conditional and the marginal version of the WAIC. I agree that for the MAE, it might not change so you can keep the equation, but maybe for [Brier score](https://scikit-survival.readthedocs.io/en/stable/api/generated/sksurv.metrics.brier_score.html#sksurv.metrics.brier_score) a new way to compute inverse probability will be used so you might don't want to write it down explicitly... 

Thus, as you point it out, it might depend on the metric, let me know if it is clear!

maylistran01: Yes it is super clear ! 
Then, would you agree if we keep the equations for the AIC, BIC, MAE, the MSE and the general R2 (used by sklearn_metrics) and not for the other metrics (WAIC, Marginal and conditional R2, Brier score, dynamic AUC)?

docs/models_evaluation.md
```diff
@@ -1,9 +1,123 @@
...
+
+## Fit metrics
+
+In leaspy, 3 negative log likelihood (nll) are stored in the model’s json file, in the `fit_metrics` part:
+- `nll_attach`: Corresponds to the nll attach to the data
```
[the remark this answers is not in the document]
maylistran01: Ok, I've added the formula.

docs/to_go_further.md
maylistran01: Ok, solved.

docs/models_evaluation.md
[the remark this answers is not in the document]
maylistran01: Ok, I've updated the issue https://github.com/aramis-lab/leaspy/issues/107.

PR #368, opened 2025-05-13
[DOC] Documentation simulate

maylistran01:
This PR Is for the documentation of simulate. 

### TODO: 
We will need to update the `regular`parameter when it has been changed in the function. 

### Question:
Should we add a link to the docstrings documentation for more details on the function?


docs/algorithms.md
```diff
@@ -17,11 +17,70 @@
...
+>>> print(simulated_data.data.to_dataframe().set_index(['ID', 'TIME']).head())
+```
+
+### Output
+
+The output is a Data object with ID, TIME and simulated values of each feature. 
+
+### Setting options
+
+Three options for simulating the visit times exist in Leaspy, which can be specified in visit_param dictionary:
```
caglayantuna, line 71:
```suggestion
There are three options to simulate the visit times in Leaspy, which can be specified in visit_param dictionary: 
```

caglayantuna (commented):
Thanks @maylistran01 ! It seems very good. I made some small suggestions also I have a question. 

I think that we mention Leaspy model as `trained model`  first time. We generally say `fitted model`. I could be wrong, we should check other documents. It would be better to be consistent.

docs/algorithms.md
```diff
@@ -17,11 +17,70 @@
 ## Estimate
 ## Simulate
 
+This section describes the procedure for simulating new patient data based on a trained Leaspy model and user-defined parameters. The simulation involves the following steps:
+
+**Step 1: Generation of Individual Parameters**
+For each simulated patient, individual parameters (tau, xi, and sources) are sampled from normal distributions defined by the model’s mean and standard deviation. These model parameters come from a previously trained Leaspy model.
+
+**Step 2: Generation of Visit Times**
```
JulietteOrtholand, line 25: Or can be provided by the user?

docs/algorithms.md
```diff
@@ -17,11 +17,70 @@
...
+Visit times are generated based on user-specified visit parameters, such as the number of visits, spacing between visits, and follow-up duration. These parameters are provided through a dictionary.
+
+**Step 3: Estimation of Observations**
+The estimate function from Leaspy is used to compute the patient observations at the generated visit times, based on the individual parameters. Gaussian noise (β-noise) is added to the observations to reflect variability.
```
JulietteOrtholand, line 29: Gaissian or Beta? Maybe here say that it is logistic associated with beta? For a linear progression I am not sure it makes sens to put a beta distribution

docs/algorithms.md
```diff
@@ -17,11 +17,70 @@
...
+    }
+>>> simulation_settings = AlgorithmSettings(
+        "simulation",
+        seed=0,
+        features=["MDS1_total", "MDS2_total", "MDS3_off_total", 'SCOPA_total', 'MOCA_total', 'REM_total', 
+                  'PUTAMEN_R', 'PUTAMEN_L', 'CAUDATE_R', 'CAUDATE_L'],
+        visit_parameters=visits_params
+    )
+>>> simulated_data = leaspy_logistic.simulate(simulation_settings)
+>>> print(simulated_data.data.to_dataframe().set_index(['ID', 'TIME']).head())
```
JulietteOrtholand, line 62: Could you add the actually printed code? as in the nutshell example

docs/algorithms.md
```diff
@@ -17,11 +17,70 @@
 ## Estimate
 ## Simulate
```
JulietteOrtholand, line 18: I am wondering to what extend you should try to make the link with how data are simulated in the litterature, equations? do document a bit the black box behind, WDYT? I am not thinking about something super complex, maybe getting inspiration of [this](https://arxiv.org/pdf/2501.08960) p7

docs/algorithms.md
```diff
@@ -17,11 +17,70 @@
...
 ## Data Generalities
 - monotonicity
 - NaN 
 - number of visits 
 - outliers
 - not enough patients 
 - parameters don't converge 
-- score don't progress
\ No newline at end of file
+- score don't progress
```
JulietteOrtholand, line 86: Necessary? Might cause futur conflicts?

docs/to_go_further.md
```diff
@@ -6,4 +6,4 @@
 ## What kind of scientific question could be answered with leaspy? 
 ### Covariates 
 Some papers have been published adressing such kind of questions [REF].
-[Small summary and an exemple of plot]
\ No newline at end of file
+[Small summary and an exemple of plot]
```
JulietteOrtholand, line 9: Necessary? Might cause futur conflicts?

JulietteOrtholand (commented):
Thank you very much for this PR! The explanation and the code so that your work can be easily used in the futur is here! I am wondering to what extend you should document a bit the "black box" process, WDYT? Anyway, nice work :)

docs/algorithms.md
```diff
@@ -17,11 +17,70 @@
...
+>>> visits_params = {
+        'patient_nb': 200,
+        'visit_type': "regular",
+        'regular_visit': 1,
+        'first_visit_mean': 0.,
+        'first_visit_std': 0.4,
+        'time_follow_up_mean': 11,
+        'time_follow_up_std': 0.5,
+        'distance_visit_mean': 2 / 12,
+        'distance_visit_std': 0.75 / 12,
```
JulietteOrtholand, line 52: Could you maybe document that in the setting options? Or put a ref to the automatically generated documentation of the function?

docs/algorithms.md
maylistran01: Ok, I've added the printed code.

maylistran01 (commented):
Thank you Juliette for the review ! 
I've made the quick-fixes, and I will answer to your first comment later.

docs/algorithms.md
NicolasGensollen: If `simulate` takes an instance of `AlgorithmSettings` as input, then it doesn't need to document it further than something like this:

```python
def simulate(settings: AlgorithmSetttings):
    """Simulate stuffs.

    Parameters
    ----------
    settings : AlgorithmSettings
        The settings for the simulation algorithm.
    """
```

The `AlgorithmSettings` will document its own attributes:

```python
class AlgorithmSettings:
    """Class representing the possible settings for an algorithm.

    Attributes
    ---------
    patient_nb : int
        The number of patients to simulate.

    visit_type : str
        The type of visits to simulate.

    ...
    """
```

Does that make sense ?

PR #371, opened 2025-05-19
Clean dirty hacks

### LINKED ISSUE #334: Clean the hacks to ease the v1 -> v2 transition

NicolasGensollen:
Closes #334


caglayantuna:
Thanks @NicolasGensollen, it looks good to me! I have only one question. Do you know why are the test files changing?

NicolasGensollen:
Thanks for the review @caglayantuna !

The results of a model fit depend on the order in which variables accessed to be used (for sampling for example).
To match the results from V1, we had to make sure the order was the same, hence the appropriately named "dirty hacks" which were hardcoding that order. When removing those, the order changed (I made sure it was deterministic by sorting) and the results also changed. This is why I had to update the references. Does that make sense ?

PR #374, opened 2025-05-20
Clean dirty hacks 2

NicolasGensollen:
Following #371 addressing forgotten bits of #334

### DIFF
--- a/tests/functional_tests/api/test_api_personalize.py
+++ b/tests/functional_tests/api/test_api_personalize.py
@@ -648,14 +648,14 @@ class LeaspyPersonalizeRobustnessDataSparsityTest(LeaspyPersonalizeTestMixin):
         self._robustness_to_data_sparsity(
             "logistic_diag_noise",
             "mode_posterior",
-            {"n_iter": 150000},
+            {"n_iter": 100},
         )
 
     def test_multivariate_logistic_diagonal_mean_posterior(self):
         self._robustness_to_data_sparsity(
             "logistic_diag_noise",
             "mean_posterior",
-            {"n_iter": 150000},
+            {"n_iter": 100},
         )
 
     def test_multivariate_logistic_diagonal_no_source_scipy_minimize(self):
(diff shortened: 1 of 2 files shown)


tests/functional_tests/api/test_api_personalize.py
```diff
@@ -648,12 +648,14 @@ def test_multivariate_logistic_diagonal_mode_posterior(self):
         self._robustness_to_data_sparsity(
             "logistic_diag_noise",
             "mode_posterior",
+            {"n_iter": 100},
```
NicolasGensollen, line 651: @JulietteOrtholand I followed your advice to play with the number of iterations. It does something, but basically the opposite of what I was expected. With large numbers of iterations (tried up to 15k iterations), the results are different enough to break the tests while they happily pass with low numbers of iterations (<= 100 iterations). The default number (1k iterations) breaks also.
Is it possible that the algorithm has multiple convergence points ? For example, I saw close results with 15k iterations except that some numbers were opposite (for xi if I remember correctly).

JulietteOrtholand: Is it the same comportment before and after your changes? 

I am not that surprised that if you don't give exactly the same data the algorithm does not converge at the same point (yet there was a [paper](https://ieeexplore.ieee.org/abstract/document/8759198) that was supposed to show that leaspy was robust to missing data...). How wide are the differences? 
A second thing is that they might be initialised at the same point. The initialisation is an approximation of the model, so they might have a really close approximation as initialisation, yet converge toward slightly different points.

What I would suggest, is to document that the comportment is the same before and after your changes (and maybe with master?) And maybe we should think about a more robust test? WDYT?

JulietteOrtholand: Could it be linked to the rank of the source in sampling? When you check the differences, the major differences at 20k for the older seem to be on the sources. I am not sure for the rational behind nevertheless... To what extent starting by something that might be hard to estimate, when all the individual parameters might be not the most appropriate for the model could have an impact? Could it be that estimating sources before tau is not a "good" idea? (they both play on temporal aspect) Could it be that starting with the sources is not a good idea? 

Maybe testing other orders could give us information on that? We could discuss that rapidly this afternoon also if you want.

NicolasGensollen:
As discussed today, let's merge and move on.

PR #376, opened 2025-05-20
Use dataset.event_time_name instead of hardcoded 'EVENT_TIME'

GabrielleCasimiro:
Replace the hardcoded string "EVENT_TIME" with the dynamic attribute dataset.event_time_name. Improves flexibility by allowing the code to work with datasets that may use different event time column names.


NicolasGensollen (approved):
Thanks @GabrielleCasimiro !

PR #328, opened 2025-03-03
[DOC] Add examples gallery

NicolasGensollen:
Closes #373

PR #378, opened 2025-05-20
extend gitignore

caglayantuna:
After the last changes (#328) related to examples in Leaspy documentary, rendering documentation creates directory and some output folders because of the fit algorithm. This PR adds these changes to .gitignore file in order to ignore them in git when we render the doc locally.


NicolasGensollen (approved):
LGTM, thanks @caglayantuna !

PR #405, opened 2025-06-02
Re-organize algorithm names

NicolasGensollen:


NicolasGensollen:
> I tried to resolve conflicts in this PR, but there are some import problems as well. Are you still working on this @NicolasGensollen?

I stopped at some point because I was unsure this would be truly useful. I'm still debating whether we should merge it or just close it. Any opinion on your side ?

PR #415, opened 2025-06-16
Covariate model - Load Data

GabrielleCasimiro:
### Description

This PR introduces the first part of the covariate model integration, focusing specifically on **data loading**.

### Changes included

- Added `covariate_dataframe_data_reader.py` to support loading covariate data and check that the covariate data provided by the user conforms to the expected format
- Updated `data.py`, `dataset.py`, `factory.py`, and `individual_data.py` to handle additional covariate inputs  

### Tested with

- Manual runs on simulated datasets with and without covariates  
- No impact expected on current functionalities if covariates are not used

### DIFF
--- a/src/leaspy/io/data/covariate_dataframe_data_reader.py
+++ b/src/leaspy/io/data/covariate_dataframe_data_reader.py
@@ -1,0 +1,244 @@
+import warnings
+
+import numpy as np
+import pandas as pd
+
+from leaspy.exceptions import LeaspyDataInputError
+from leaspy.utils.typing import Dict, FeatureType, IDType, List, Optional
+
+from .abstract_dataframe_data_reader import AbstractDataframeDataReader
+from .individual_data import IndividualData
+from .visit_dataframe_data_reader import VisitDataframeDataReader
+
+__all__ = ["CovariateDataframeDataReader"]
+
+
+class CovariateDataframeDataReader(AbstractDataframeDataReader):
+    """
+    Methods to convert :class:`pandas.DataFrame` to `Leaspy`-compliant data containers for longitudinal data with covariates.
+
+    Parameters
+    ----------
+    covariate_names: List[str]
+        Names of the columns in dataframe that contains the covariates
+
+    Raises
+    ------
+    :exc:`.LeaspyDataInputError`
+    """
+
+    def __init__(
+        self,
+        *,
+        covariate_names: List[str],
+    ):
+        super().__init__()
+        if not covariate_names:
+            raise LeaspyDataInputError("You must prrovide at least one covariate name.")
+        self.covariate_names = covariate_names
+        self.visit_reader = VisitDataframeDataReader()
+
+    @property
+    def long_outcome_names(self) -> List[FeatureType]:
+        """Name of the longitudinal outcomes in dataset"""
+        return self.visit_reader.long_outcome_names
+
+    @property
+    def n_visits(self) -> int:
+        """Number of visit in the dataset"""
+        return self.visit_reader.n_visits
+
+    ######################################################
+    #               COVARIATE METHODS
+    ######################################################
+
+    def _check_headers(self, columns: List[str]) -> None:
+        """
+        Check mendatory dataframe headers
+
+        Parameters
+        ----------
+        columns: List[str]
+            Names of the columns headers of the dataframe that contains patients information
+        """
+        self.visit_reader._check_headers(columns)
+
+    def _set_index(self, df: pd.DataFrame) -> pd.DataFrame:
+        """
+        Set the index suited for the type of information contained in the dataframe
+
+        Parameters
+        ----------
+        df: pd.DataFrame
+            Dataframe with patient information
+
+        Returns
+        -------
+        df: pd.DataFrame
+            Dataframe with the right index
+        """
+
+        return self.visit_reader._set_index(df)
+
+    def _clean_dataframe_covariates(
+        self, df: pd.DataFrame, *, drop_full_nan: bool, warn_empty_column: bool
+    ) -> pd.DataFrame:
+        """
+        Clean the dataframe that contains patient information
+
+        Parameters
+        ----------
+        df: pd.DataFrame
+            Dataframe with patient information
+
+        drop_full_nan: bool
+            If set to True, raw full of nan are dropped
+
+        warn_empty_column: bool
+            If set to True, a warning is raise for columns full of nan
+
+
+        Returns
+        -------
+        df: pd.DataFrame
+            Dataframe with clean information
+        """
+
+        # [SPECIFIC] check_available_data
+        df_covariate = df.copy(deep=True)
+
+        # Assert covariates columns are the only one available
+        assert (df_covariate.columns == self.covariate_names).all()
+
+        # Check if there are missing values (NaN) in the covariate
+        for covariate in self.covariate_names:
+            if df_covariate[covariate].isna().any():
+                raise LeaspyDataInputError(
+                    f"Covariate '{covariate}' contains missing values (NaN)."
+                    "Please ensure that values are provided for each visit."
+                )
+
+        # Check covariate good format
+        for covariate in self.covariate_names:
+            if not np.array_equal(
+                df_covariate[covariate], df_covariate[covariate].astype(int)
+            ):
+                raise LeaspyDataInputError(
+                    f"Covariate '{covariate}' must contain only integer values."
+                )
+            df_covariate[covariate] = df_covariate[covariate].astype(int)
+
+        # Assert one unique covariate per patient and group to drop duplicates
+        if (
+            not (df_covariate.groupby("ID").nunique()[self.covariate_names].eq(1))
+            .all()
+            .all()
+        ):
+            raise LeaspyDataInputError(
+                "There must be only an unique covariate value per patient."
+            )
+        df_covariate = df_covariate.groupby("ID").first()
+
+        # Covariate must be empty to raise an error
+        if len(df_covariate) == 0:
+            raise LeaspyDataInputError("Dataframe should have at least 1 covariate")
+
+        # Assert at least 2 different values per covariate
+        for covariate in self.covariate_names:
+            n_value = df_covariate[covariate].nunique(dropna=False)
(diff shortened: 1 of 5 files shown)


src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+        Returns
+        -------
+        df: pd.DataFrame
+            Dataframe with clean information
+        """
+
+        # [SPECIFIC] check_available_data
+        df_covariate = df.copy(deep=True)
+
+        # Assert covariates columns are the only one available
```
NicolasGensollen, line 111: You don't need all those comments, the code is clear enough and speaks for itself. If you comment everything, the file will be twice as long and there will be a lot of noise for the reader. Save comments for things that are not obvious when reading the code.

```suggestion
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+        assert (df_covariate.columns == self.covariate_names).all()
```
NicolasGensollen, line 112: Usually, we use assertions in tests and prefer proper error raising in code because it gives much more details to the end user. So I'd suggest something like that:

```suggestion
        if not (df_covariate.columns == self.covariate_names).all():
            raise LeaspyDataInputError(
                f"The covariate column names {df_covariate.columns} are "
                f"different from the provided covariate names {self.covariate_names}."
            )
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # Check if there are missing values (NaN) in the covariate
```
NicolasGensollen, line 114:
```suggestion
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+            If set to True, a warning is raise for columns full of nan
+
+
+        Returns
+        -------
+        df: pd.DataFrame
+            Dataframe with clean information
+        """
+
+        # [SPECIFIC] check_available_data
```
NicolasGensollen, line 108:
```suggestion
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # Check if there are missing values (NaN) in the covariate
+        for covariate in self.covariate_names:
+            if df_covariate[covariate].isna().any():
+                raise LeaspyDataInputError(
+                    f"Covariate '{covariate}' contains missing values (NaN)."
+                    "Please ensure that values are provided for each visit."
+                )
+
+        # Check covariate good format
```
NicolasGensollen, line 122:
```suggestion
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+            not (df_covariate.groupby("ID").nunique()[self.covariate_names].eq(1))
+            .all()
+            .all()
+        ):
+            raise LeaspyDataInputError(
+                "There must be only an unique covariate value per patient."
+            )
+        df_covariate = df_covariate.groupby("ID").first()
+
+        # Covariate must be empty to raise an error
```
NicolasGensollen, line 143:
```suggestion
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+        if len(df_covariate) == 0:
+            raise LeaspyDataInputError("Dataframe should have at least 1 covariate")
+
+        # Assert at least 2 different values per covariate
+        for covariate in self.covariate_names:
+            n_value = df_covariate[covariate].nunique(dropna=False)
+            if n_value < 2:
```
NicolasGensollen, line 150: You can use the walrus operator to simplify a bit:

```python
if (n_value := df_covariate[covariate].nunique(dropna=False)) < 2:
    raise LeaspyDataInputError(...)
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+            If set to True, a warning is raise for columns full of nan
+
+
+        Returns
+        -------
+        df: pd.DataFrame
+            Dataframe with clean information
+        """
+
+        # Check visits
```
NicolasGensollen, line 182:
```suggestion
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+        df_visit = self.visit_reader._clean_dataframe(
+            df.drop(columns=self.covariate_names),
+            drop_full_nan=drop_full_nan,
+            warn_empty_column=warn_empty_column,
+        )
+
+        # Check covariates
```
NicolasGensollen, line 189:
```suggestion
```

src/leaspy/io/data/covariate_dataframe_data_reader.py
```diff
@@ -0,0 +1,245 @@
...
+        df_covariate = self._clean_dataframe_covariates(
+            df.reset_index()
+            .drop(self.long_outcome_names + ["TIME"], axis=1)
+            .set_index("ID"),
+            drop_full_nan=drop_full_nan,
+            warn_empty_column=warn_empty_column,
+        )
+
+        # [SPECIFIC] prepare_clean_output
```
NicolasGensollen, line 198:
```suggestion
```

src/leaspy/io/data/data.py
```diff
@@ -59,7 +59,8 @@ def __init__(self):
         self.event_time_name: Optional[str] = None
         self.event_bool_name: Optional[str] = None
 
-        # Cofactor information (?)
+        # Covariate information
```
NicolasGensollen, line 62:
```suggestion
```

src/leaspy/io/data/data.py
```diff
@@ -529,25 +541,38 @@ def from_individual_values(
             if not event_bool_name or not event_time or not event_bool:
                 raise ("Not coherent inputs for longitudinal data")
 
+        # Covariates input checks
```
NicolasGensollen, line 544:
```suggestion
```

src/leaspy/io/data/data.py
```diff
@@ -529,25 +541,38 @@ def from_individual_values(
...
+        if not covariate_names:
```
NicolasGensollen, line 545: Small trick to simplify these conditions.
Also, please raise a proper error with a message as helpful as possible for the user:

```python
if (covariate_names is None) == (covariates is None):
    raise ValueError(
        "Not coherent inputs for covariate data: \n "
        f"covariate_names = {covariate_names} and \n "
        f"covariates = {covariates}."
    )
```

src/leaspy/io/data/data.py
```diff
@@ -529,25 +541,38 @@ def from_individual_values(
...
                 indiv.add_event(event_time[i], event_bool[i])
+            if covariate_names:
+                indiv.add_covariates(covariates[i])
             individuals.append(indiv)
 
         return Data.from_individuals(
             individuals, headers, event_time_name, event_bool_name
         )
 
+        # Covariate input checks
```
NicolasGensollen, line 567:
```suggestion
```

src/leaspy/io/data/dataset.py
```diff
@@ -105,7 +105,9 @@ def __init__(self, data: Data, *, no_warning: bool = False):
         self.event_time: Optional[torch.FloatTensor] = None
         self.event_bool: Optional[torch.IntTensor] = None
 
-        # Cofactor information (?)
+        # Covariate information
```
NicolasGensollen, line 108:
```suggestion
```

NicolasGensollen (commented):
Thanks for working on this @GabrielleCasimiro !

I made a quick first pass. The code looks good to me. I made a few suggestions, mainly on removing useless comments.

src/leaspy/io/data/data.py
GabrielleCasimiro: We kept it for Event and Longitudinal input checks. If I remove it for Covariates input checks, should I remove the other ones as well ?

src/leaspy/io/data/dataset.py
GabrielleCasimiro: Same question here, as we kept the Patients, Longitudinal outcome and Event information comments. Should I remove all of them of keep the comment for Covariate info ?

src/leaspy/io/data/data.py
NicolasGensollen: When in doubt, it's always useful to open an the interpreter and try:

```python
>>> def f(x, y):
...     if (x is None) == (y is None):
...         print (f"x = {x}, y = {y}")
...     else:
...         raise ValueError("bad bad bad")
...
>>> f(1, 2)
x = 1, y = 2
>>> f(None, None)
x = None, y = None
>>> f(1, None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 5, in f
ValueError: bad bad bad
>>> f(None, 1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 5, in f
ValueError: bad bad bad
```

Is that the desired behavior ?

GabrielleCasimiro: Thanks for the tip. I only want an `if`condition (not `if` `else`) so I prefer this form :
```python
if (covariate_names is None) != (covariates is None):
    raise ValueError("Not coherent inputs for covariate data: \n "
        f"covariate_names = {covariate_names} and \n "
        f"covariates = {covariates}.")
```
So that the error is raised when one is `None`and the other is not.

NicolasGensollen: I misunderstood your question, sorry about that.
Yes, you should totally do that 👍

PR #419, opened 2025-06-26
Clean docstrings io.outputs.result.py

KaisaridiSofia:
Add and complete the docstings in io.outputs.result.py as mentioned in issue #122

### DIFF
--- a/src/leaspy/io/outputs/result.py
+++ b/src/leaspy/io/outputs/result.py
@@ -28,7 +28,8 @@ class Result:
     Parameters
     ----------
     data : :class:`.Data`
-        Object containing the idx, time-points and observations of the patients
+        Object containing the information of the individuals, 
+        in particular the time-points :math:`(t_{i,j})` and the observations :math:`(y_{i,j})`.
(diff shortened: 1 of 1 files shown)


src/leaspy/io/outputs/result.py
```diff
@@ -29,23 +29,23 @@ class Result:
     ----------
     data : :class:`.Data`
         Object containing the idx, time-points and observations of the patients
```
caglayantuna, line 31: I know that you didn't write this but it would be nice to improve it now. I found a better version [here](https://github.com/aramis-lab/leaspy/blob/v2/src/leaspy/models/base.py#L177).  Maybe it is not exactly the same because we don't have option here (Data, Dataset), I guess. In any case, we can improve the text by using `information of the individuals` rather than `idx` at least. What do you think?

KaisaridiSofia: Nice remark! I updated it.

caglayantuna:
Thanks @KaisaridiSofia . You need to rebase your branch with the latest v2 before merging it.

PR #441, opened 2025-07-16
Cleaning Docstring constant.py

lea-aguilhon:
Add and complete the docstrings in model/constant.py as mentioned in issue https://github.com/aramis-lab/leaspy/issues/122.


caglayantuna (commented):
thanks @lea-aguilhon . In line 38, link is not up-to-date. It should be     :class:`~leaspy.algo.personalize.constant_prediction_algo.ConstantPredictionAlgorithm`.

PR #450, opened 2025-08-01
Documentation logistic 

maylistran01:
Here is the PR for the documentation for the logistic model. 
I followed the plan used by @lea-aguilhon in the joint model part.

Thanks in advance for the review !


caglayantuna:
Thanks @maylistran01 , it seems ok to me, merging.

PR #483, opened 2026-01-27
Fixing: Doc Warnings

MendezSebastianP:
- We will solve as much as warnings/errors as we can. We start with about 320~ : 290~ warnings, 30 errors. 
- There are too many errors in various files, even when we try to fix a group of them, LLMs are not that performant fixing them, most of the errors will be solved by hand...
- As some errors are located in the build generated files (e.g. `.rst`), we can not be always sure about where these errors come from. And as the build time from the doc is about 5 minutes, sometimes the fixes are not very accurate. **That means that we are obligated to modify more lines/files that what we really need**.
- In this first commit we solved all errors and some warnings, ~268 left.


MendezSebastianP:
Most of the changes are related with the warnings, but some others no. To check if a warning was fixed i had to run `make doc` which cleans and build the doc from scratch, it takes ~3 minutes. So fo each 'run', i changed different files/lines and after i checked, so if it solve some warnings (but the rest of the chamges did nothing) i still push the changes. I could not do +300 builds, even if i did +50. Some changes could so potentially change undesired descriptions.

MendezSebastianP:
@KaisaridiSofia , @GabrielleCasimiro 
When you have some time, could you review the [doc](https://leaspy--483.org.readthedocs.build/en/483/) and confirm everything looks OK? The most important part is checking the links in functions/classes/references  in API reference, links in the doc, and corrupted/not-well-rendered parts. We had WAY more broken links than we thought in the documentation/docstrings. Normally these changes help to fix this and have more accurate redirections, but i have not check everything.

Issue #487, opened 2026-03-06
Add new features for the mixture model


KaisaridiSofia:
Integrate the functions demonstrated in the [example](https://leaspy.readthedocs.io/en/latest/auto_examples/plot_04_mixture.html) for the extraction of individual parameters and the plotting in the main code.

- [x] extraction of individual parameters and probabilities
- [x] plotting

Issue #486, opened 2026-03-06
Fix personalization for mixture model


KaisaridiSofia:
There is a problem with the personalization algorithm and the mixture model because of the extra dimension of the model parameters.
To ensure the algorithm works as it is supposed we should add the following code in `mixture.py` in line 422, in the function `put_individual_parameters`.

```
for k in ["xi", "tau"]:
    if state[k].ndim != 2:
        state[k] = state[k].reshape(-1, 1)
```

This forces a reshaping in tau and xi in a way that they can then be used accordingly by the personalization algorithm. It works globally for any number of clusters as it controls if the number of dimensions and not the dimension of the variables.

KaisaridiSofia:
Resolved in PR #489

PR #489, opened 2026-03-10
Remaining tasks for the mixture model

KaisaridiSofia:
This PR is intended to implement remaining details for the mixture model.

- [x]  issue #486 
- [x]  issue #487. Specifically we added functions to:
   - [x] plot average trajectory per cluster
   - [x] extract the individual parameters
   - [x] plot patient observations
   - [x] plot individual trajectories
   - [x] ensure convergence plots

### DIFF
--- a/src/leaspy/algo/fit/fit_output_manager.py
+++ b/src/leaspy/algo/fit/fit_output_manager.py
@@ -13,7 +13,7 @@ from matplotlib.backends.backend_pdf import PdfPages
 from matplotlib.lines import Line2D
 
 from leaspy.io.data import Dataset
-from leaspy.models import McmcSaemCompatibleModel
+from leaspy.models import McmcSaemCompatibleModel, LogisticMultivariateMixtureModel
 
 from .base import FitAlgorithm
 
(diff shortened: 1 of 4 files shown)


KaisaridiSofia:
while checking the fit output manager, I managed to correct the error for the convergence plots for the joint model as well, mentioned in issue #491

MendezSebastianP:
@KaisaridiSofia 
1. Do you have any data+scripts to test the new features ?
2. Do you know why there is "@doc_with_super()" in some of your classes (mixture)?

Other than that, i checked the changes in the code and they seem good. We just have to open an issue where we talk about the unity tests, because there is no tests for mixture and ordinal. At some point i will have to check how that works and we will add some, to make sure that if someone touches something it wont break mixture models, in that sens it is a bit vulnerable.

KaisaridiSofia:
Thanks @MendezSebastianP .

For the "@doc_with_super()" I am not sure for its use or even if it is necessary. There is a possibility that it's been left there from the passing from leaspy v1 to v2. I think I will open soon a new PR to refactor the mixture model and see if it can be merged to logistic and remove any unnecessary things.

You are right about the tests. Can you open a new issue to discuss it?

MendezSebastianP (commented):
Everything is ok

Issue #439, opened 2025-07-11
Clarify the use of the "name" parameter in model definition


lea-aguilhon:
**Context:**
During discussions about model definition, concerns were raised regarding the use of the `name` parameter when initializing models. For 
example:
`model_logistic = LogisticModel(name="model_logi", dimension= X, sources =X)`

The current implementation uses this name parameter in the ScipyMinimizeAlgorithm (algo/personalize/base.py), which may not be the most intuitive approach.

**Current implementation:**

- The name parameter is defined in  `model/classBase`
- It's used in `algo/personalize/base.py` in class `ScipyMinimizeAlgorithm`

The parameter is currently used to determine model behavior, where it might be more appropriate to use the model type (Logistic, Joint, etc.)

**Specific Use Cases:**

` if "joint" in model.name:`
`            data_type = "joint"`
`           factory_kws = {"nb_events": model.nb_events}`
`        else:`
`            data_type = "visit"`
`            factory_kws = {}`

and here:
`        if with_jac and not self.is_jacobian_implemented(model):`  
`              warnings.warn(`  
 `                 "In `scipy_minimize` you requested `use_jacobian=True` but it "`  
`                  f"is not implemented in your model {model.name}. "`  
`                  "Falling back to `use_jacobian=False`..."`  
`              )`

KaisaridiSofia:
Generally when we need to save and then reload a model, that doesn't work unless we name the model with the specific name of the instance declared in the model factory. There are some errors triggered that sent us back to `models/base.py`.

After a quick look at the source file I think we should change it with another argument, like `model.type` that suggests the instance of the model and leave the user to give different names to his/her model.

KaisaridiSofia:
I found an easy fix for now, to suppress the user-given argument at least to avoid the problems in `save` and `load` functions. 

The argument `name` is needed in the factory and the `ModelInterface`so I think we cannot bypass it completely. We can give by default in each model class the name needed to proceed correctly. In that way the user cannot give a proper name on his/her model but he/she can still differentiate like this:

```
my_path = Path("/Users/sofia.kaisaridi/leaspy_v2_sk/test_model_name")

model_multi = LogisticMultivariateMixtureModel(
    source_dimension=2, dimension=6, n_clusters=3,
    obs_models="gaussian-diagonal",
)

model_multi.fit(leaspy_dataset, "mcmc_saem", seed=1312, n_iter=500, 
[...]

Issue #344, opened 2025-03-26
Problem with total likelihood


KaisaridiSofia:
In the code in [abstract_model](https://github.com/aramis-lab/leaspy/blob/d17a841b160dcbf2aa9e8c3de838ca99b684d7ba/src/leaspy/models/abstract_model.py#L785) we compute : 

for k in ("nll_attach", "nll_regul_ind_sum"):
            suff_stats[k] = state[k]
        suff_stats["nll_tot"] = (
            suff_stats["nll_attach"] + suff_stats["nll_regul_ind_sum"]
        )

**1st issue :** At the end of the fit it does not appear in the state, we can not print it. We have the same problem also with : 

- "nll_regul_pop_sum" (calculated as : "nll_regul_log_g" + "nll_regul_log_v0" + "nll_regul_betas"
- "nll_regul_all_sum" (calculated as "nll_regul_pop_sum" + nll_regul_ind_sum)

the terms inside the parentheses are well calculated in the end

**2nd issue :** It's not calculated correctly. It should also add the term "nll_regul_pop_sum"

KaisaridiSofia:
It was fixed in the PR #494 . The most important changes are in the source files `variables/specs.py` for the correct calculations, `variables/dag.py` for the sorting of the ancestors and in `models/mcmc_saem_compatible.py`for the correct calculation of the nll_tot.

The original idea behind the exclusion of the term nll_regul_pop was that since we do not have priors on the population parameters, the term does not really changes a lot during the fit. Nevertheless since the nll is mainly used for model comparison it should be included in the total sum. Also if we want to explore a fully bayesian leaspy in the future, this will be crucial.

So after the merge of the dedicated PR, the result of the total likelihood for the models fitted with leaspy will be different, but this will not affect the convergence as the quantity `nll_regul_pop_sum` is essentially added in all the iterations and has minimum fluctuations during the fit.

This created a lot of errors from the tests are the usecase examples expected also the terms `nll_regul_pop_sum` and `nll_regul_all_sum` in the `fit_metrics` on the .json files. So I added some lines to exclude these in `algo/fit/base.py` for now. We can retrieve them later but first we have to rerun all the examples used in the tests to get the results of these metrics as well. For now we can print the terms at the end of the fit, passing by the state: `mode.state["nll_regul_pop_sum"]`, `mode.state["nll_regul_all_sum"]`

KaisaridiSofia:
[...]

Issue #121, opened 2025-02-17
Scalar gaussian noise for a multivariate model


NicolasGensollen:
In GitLab by @sofia.kaisaridi on Feb 17, 2025, 12:08

We let the user to choose between a "gaussian-diagonal" and a "gaussian-scalar" observational model when implementing a multivariate model. The difference between the two methods is that the "diagonal" option adds a noise for each feature (1\<=k\<=d) corresponding to : 

undefined{width=453 height=43}

while the "scalar" adds a global noise shared between all the features, corresponding to :

undefined{width=453 height=43}

My question is, do we really need to keep the scalar option for a multivariate model? I thought that the basic formulation for the DCM uses the diagonal noise.

NicolasGensollen:
In GitLab by @sofia.kaisaridi on Feb 21, 2025, 14:57

If we choose to keep the scalar noise as an option for a multivariate model, I think we should change the default method to add a diagonal noise when the user tries to fit a multivariate model.

Now if we use a `leaspy_data` with multiple features and just do :

`leaspy = Leaspy('logistic')`

`leaspy.fit(leaspy_data, settings=algo_settings)` 

we get one noise for all the features:

/Users/sofia.kaisaridi/leaspy/src/leaspy/models/abstract_multivariate_model.py:180: UserWarning: You did not provide \`source_dimension\` hyperparameter for multivariate model, setting it to ⌊√dimension⌋ = 3. warnings.warn(

|--------------------------------------------------| 0/100 iterations

/Users/sofia.kaisaridi/Library/Python/3.10/lib/python/site-packages/torch/\__init_\_.py:696: UserWarning: torch.set_default_tensor_type() is deprecated as of PyTorch 2.1, please use torch.set_default_dtype() and torch.set_default_device() as alternatives. (Triggered internally at /Users/runner/work/pytorch/pytorch/pytorch/torch/csrc/tensor/python_tensor.cpp:453.) \_C.\_set_default_tensor_type(t)

undefined{width=813 height=382}

undefined{width=813 height=464}

KaisaridiSofia:
If we precise the `dimension`(number of features) the observation model is gaussian diagonal. We need to enforce this even when the user does not give a dimension.

KaisaridiSofia:
I worked on a fix that makes gaussian-diagonal the default observation model.
Then an error is raised asking for the `dimension` if it is not defined that says: 

[...]

PR #494, opened 2026-03-23
Bug fixes

KaisaridiSofia:
With this PR we intend to fix various bugs that need to be treated for the release v2.1

- [x] issue #439 
- [x] issue #121 
- [x] issue #344


MendezSebastianP:
Nice job there! could you please tell me which are the commits linked with each issue? the most relevant. Thanks

KaisaridiSofia:
Here is a list of the commits:

- [4b210fa](https://github.com/aramis-lab/leaspy/commit/4b210fa7bfe7f54321e469496105a369f86c0fa8) is for the issue #439 
- [e3f8878](https://github.com/aramis-lab/leaspy/commit/e3f887832bf54c0806fec617463d1f17dddde699), [f0b30ec](https://github.com/aramis-lab/leaspy/commit/f0b30eca7c51d99e17aea09640ce195a27294e2b) and [ea7ba60](https://github.com/aramis-lab/leaspy/commit/ea7ba60f49dbc09d77ff5bfc2f8f352fcee7b4a3) are for the issue #121 
   - [85a3e69](https://github.com/aramis-lab/leaspy/commit/85a3e6993f2e8658b5059d1a4148f6429e4e4edb), [85aa965](https://github.com/aramis-lab/leaspy/commit/85aa965df5c161c3cef2859cc2264ebbfcd3b7e4), [4f766e4](https://github.com/aramis-lab/leaspy/commit/4f766e4150e7be08a005345a1f80002e3e50d956), [bc8c16b](https://github.com/aramis-lab/leaspy/commit/bc8c16b135360d132afde142a6a0d97ff3bfe0e8), [f8ea356](https://github.com/aramis-lab/leaspy/commit/f8ea3569aa2d504c97b78d1362ac94281a881130), [7b9eea7](https://github.com/aramis-lab/leaspy/commit/7b9eea7b8c68225bdde724a124a929fcbf9fd0e2) are one-line changes for the **notebook** of the joint model for the same argument (obs_mdoels), I think you can check only the last one. I didn't have any other way to check how it works. There is one change in the models/joint.py at some point but I reverted it in a following PR. 
- [397d228](https://github.com/aramis-lab/leaspy/commit/397d2287b14f2e02c683c6a70550b4291d63052f) and [d241d8a](https://github.com/aramis-lab/leaspy/commit/d241d8ac7105ebc6af54e0788539259abb011e32) are for the issue #344 
   - in [20f9f57](https://github.com/aramis-lab/leaspy/commit/20f9f57efe5526b985d27bb41ea76971f8dcf1c3) there are also two source files changing but the majority of the changes concern the updated json files with the outputs
   - [1ea0d2c](https://github.com/aramis-lab/leaspy/commit/1ea0d2cdc4c8c0c5b9cbb08f619ef9cf3bbf3278) has also some changes in the json files

MendezSebastianP:
Perfect, thank you. Keep me updated about the tests issue, if you are going to modify the gold standar, if you struggle with something or if you need assistance.

KaisaridiSofia:
Yes in fact, in yesterday's meeting we said we should add the `nll_regul_pop_sum` in the output in the json files. Do you know if we can run all the models in the tests with a script and have their result json ? Or do we need to run it one by one, by hand ? If you have some time, I'd appreciate it if you investigate this a bit.

MendezSebastianP:
Yes, you right! i will check this soon, i will try to generate the json files, so i'll try to work directly in this branch.

MendezSebastianP:
The gold standard JSON files and the generation pipeline have been updated. The non _arm files currently contain approximated values for nll_regul_pop_sum (derived from a rounded nll_tot), which required loose tolerances to pass CI on Ubuntu. To fix this properly, the non-_arm files need to be regenerated on a native x86/Linux machine using MODIFY_GOLD_STANDARD = True. This will be done next week when i will work on this in my personal computer (with Ubuntu), that will be the occasion to verify if the JSON files generation is still ok in linux with all these changes, as all the test and work have be done in Mac OS. 

Even if the tests passes it must to be done.

MendezSebastianP:
in summary: 

* i wanted to add `nll_regul_pop_sum` to the JSONS, ubuntu tests crashed because i did not added `nll_regul_pop_sum` to Ubuntus JSONS

* i added ``nll_regul_pop_sum` to ubuntus JSONS, but they stll crashed because the tests have change from the last generations + after adding the new parameter

* I put more tolerance to the errors to check if that is enough to pass, they all passed, it was peding to run generate the files in Ubuntu

* I generated the files in Ubuntu, new files where generated (from scratch, not modified), because the last time they were ran with Ubuntu was long time ago. I had issues with the GPU tests (as i have a NVIDIA GPU), i update the test file and the tests passed

* Now we have new files, we can put back the atol to a more strict one.

MendezSebastianP:
There are still some work to do concerning the tests, as 11 test fail the general pytest... all theses tests are related to the GPU, There are many issues, legacy code and bugs inside. This is not the moment to adress these problems, but we need to do it after updating the test and creating a guide for the unity tests.

KaisaridiSofia:
After our discussion @MendezSebastianP , I pushed the last commit that I tried. 
Short summary of the problem: we are obliged to set the dimension parameter when we load the model with a diagonal noise (default). Ideally we would like the model to read the dimension(number of features) itself, the moment of the fit. The idea was to set a random dimension number the moment of the load and try to retrieve the info from the dataset before the fit starts.

There I try to set a default dimension to 1000, and I try to run a generic multivariate logistic model to see where the errors come from and try to overwrite the dimension argument there. I didn't manage to understand how to do that. If you have any insight, you are welcome. Otherwise we can leave it there and try to fix it in a future release.

PR #496, opened 2026-03-30
[DOCUMENTATION] Modify example from simulate 

maylistran01:
Small changes in the documentation for simulate function.


MendezSebastianP (commented):
Everything seems ok to me.

PR #499, opened 2026-04-28
Add Simulation Algo for Joint Model

JeanJNMV:
# PR #499 - Joint Simulation Algorithm 
This PR introduces the full implementation of the simulation algorithm for the joint model. The new features are:
- `JointSimulationAlgorithm` which inherits from `SimulationAlgorithm`.
- Added unit tests for `JointSimulationAlgorithm`.
- Adapted `estimate()` to work with the joint model and properly handle the events. It is used in `JointSimulationAlgorithm`.
- Added functional tests for this new `estimate`.
- `Plotting.patient_trajectories` extended to also plot the event time as a vertical line.
- `simulated_data_for_joint.csv` can be loaded with `load_dataset` (TBD).

- Fixed a bug where `SimulationAlgorithm` (for the logistic model) was not working when `source_dimension=0`.

Notes: The joint simulation algorithm handles competing events but for now, all tests have been conducted with nb_events=1.


maylistran01 (commented):
Really nice job Jean, thanks ! 

I have no additional comments on the code beyond what we've already discussed in our meetings.
One remark though, should we update the user guide "section 4.4. Simulate", so that it takes into account that we can now simulate event data?

PR #503, opened 2026-06-02
Hotfixes pre v2.1 release

MendezSebastianP:


MendezSebastianP:
6 doc warnings fixed

MendezSebastianP:
## Issue
Fixing wrong warning issue in `quick_examples`:
> The parameters ['overwrite_logs_folder', 'save_periodicity', 'plot_periodicity'] were not present by default and are likely to be unsupported.

this is completely wrong, these arguments here are fully supported, so the problem coms from `_get_algorithm` method in `BaseModel` class. Right now it process the arguments in this way:
1. it takes the kwargs (keywords arguments), and check if they are algorithm arguments in `AlgoritmSettings`, as it sees that  `['overwrite_logs_folder', 'save_periodicity', 'plot_periodicity'] ` are not, it tell us "probably this is not supported".
2. then then we give once again all kwarg, this time through `settings.set_logs`, it will check if an argument is a setting argument and it will save it, so its here where we used, and when the previous warning becomes wrong.

## The fix:
Instead of giving twice the wholle set of kwards, we give to `settings.set_logs` the settings arguments that exist, and then to  `AlgoritmSettings` the rest, so this time, the warning will be use, but for a fully unknown argument.

We also add a path to save the periodicity in arguments to avoid the second argument.

We can keep the thirdone.

PR #508, opened 2026-07-31
Add covariate model (and Schiratti's version of the logistic multivariate model)

GabrielleCasimiro:
## Summary

This PR introduces `CovariateLogisticModel`, an extension of Leaspy's multivariate logistic model that incorporates covariate effects on population parameters. As a prerequisite, it also introduces `LogisticModelSchiratti`, which follows the original parametrization from Schiratti et al. (2017).

## Motivation

The goal is to model how covariates modulate disease trajectories, specifically their effect on onset time (t0) and progression speed/shape (g, v).

## Approach

- Covariate selection is handled via a Bernoulli mask (γ).
- Covariate effects (δ) are conditioned on γ, with a prior that centers δ on 0 when γ = 0.

## Files changed

- **`LogisticModelSchiratti`** (new): reintroduces the original parametrization from Schiratti et al. (2017), separating the population parameter t0 (centered on `t0_mean`, a model parameter to be estimated) from the individual random effect tau (centered on 0, whose variance is estimated). This required new files: `logistic_Schiratti.py`, `time_reparametrized_Schiratti.py`, and `riemannian_manifold_Schiratti.py`.
- **`CovariateLogisticModel`** (new): builds on the same structure as `LogisticModelSchiratti`, but with its own set of files (`covariate_logistic.py`, `covariate_riemannian_manifold.py`, `covariate_time_reparametrized.py`), inheriting from `McmcSaemCompatibleModel`.
- **`covariate_dataframe_data_reader.py`** (modified): adds checks on the covariates provided by the user to ensure identifiability conditions are met.
- other functions required for the covariate model (Bernoulli sampler, sufficient statistics, etc.).

## Validation

The model has been validated on simulated data under controlled conditions across multiple scenarios (single/multiple covariates, disjoint vs. overlapping effects), checking recovery of ground-truth parameters, statistical power, and type-1 error. **Validation currently covers binary covariates only.**


GabrielleCasimiro:
**Note:** I fixed the typo "riemanian" → "riemannian" in model naming across several docs. I believe I caught all occurrences, but since this naming may appear in other models beyond logistic/covariate, it would be worth double-checking that nothing else in the codebase (or at runtime for the other models) still relies on the old spelling.

GabrielleCasimiro:
## `SimulationCovariateAlgorithm` — simulation for the covariate model

### Structure

`SimulationCovariateAlgorithm` (`algo/simulate/simulate_covariate.py`) is a **standalone** class, inheriting directly from `BaseSimulationAlgorithm` — it does **not** inherit from `SimulationAlgorithm` (the classic `simulate`).

It copies the parts shared with `SimulationAlgorithm` (parameter validation, visit generation, noise handling and de-duplication of too-close visits, etc.) rather than inheriting them, so the two classes can be read and compared independently without following an inheritance chain. The differences from `SimulationAlgorithm` are therefore localized and explicit:
- model type check (`CovariateLogisticModel` instead of `LogisticModel`);
- simulation of covariates per individual;
- covariate effects (`delta_t0`, `delta_g`, `delta_v0`, masked by `gamma_t0`/`gamma_g`/`gamma_v0`) taken into account when computing the simulated trajectories.

### Simulating covariates

Two mutually exclusive options, via the algorithm's parameters:

- **`covariates`**: a `DataFrame` provided directly by the user, one row per simulated patient.
- **`covariate_specs`**: a list of dictionaries, one per covariate, declaring its **type**:
  - `{"name": ..., "type": "binary", "prob": ...}` — Bernoulli draw;
  - `{"name": ..., "type": "discrete", "values": [...], "probs": [...]}` — draw from a fixed support (low-cardinality covariate, integer or standardized);
  - `{"name": ..., "type": "continuous", "mean": ..., "std": ...}` — Normal draw.

  For each type, the parameters (`prob` / `values`+`probs` / `mean`+`std`) can be **omitted**: in that case, they are automatically fetched from the covariate's empirical distribution in the model's training dataset (`model.dataset.covariates`), instead of being chosen manually.

### Current limitation

Unlike classic `simulate.py`, which relies on `model.estimate(...)` to compute simulated trajectories, `simulate_covariate.py` can't use it: `estimate()` doesn't support covariates yet.

**Future work**: extend `estimate()` to support covariates, then simplify `simulate_covariate.py` to call it directly, the same way classic `simulate.py` does.

GabrielleCasimiro:
## Remaining work on the covariate model

A few known limitations/TODOs before this can be considered fully complete:

- **`gamma` is not persisted (json / state)**: it currently lives only in `model_parameters`, so interpreting a fitted covariate model requires digging it out manually from there. As a consequence, saving/loading a model does not round-trip the `gamma` values. We need to add `gamma` to the model's saved state/json.

- **Noise type mismatch between initialization and fit is not fully guarded**:
  - Initializing a `diagonal_noise` covariate model from a `scalar_noise` logistic fit currently works silently (it's handled in the code), but should emit a **warning** since it's a lossy/implicit conversion.
  - Initializing a `scalar_noise` covariate model from a `diagonal_noise` fit is expected to crash, but currently does so without a clear message. We should add an explicit **error message** to make the failure mode understandable.

- **`personalize` and `estimate` are not yet adapted to the covariate model.** Once `estimate` supports it, `simulate` should be updated to rely on `estimate` internally instead of its current standalone covariate-specific implementation.

- **Initialization**: the fit should include a LogisticSchiratti initialization parameter by default, using the 10% of the iterations the user set. We should include an argument that control if the user wants this, lile `LogisticSchiratti_initialization'`= `True` by default. Also another one to shet the number of iterations like `init_n_iter` = `iter*0.1` by default, so the user can set the number they want. Obviously we have to set a protection in case `LogisticSchiratti_initialization` = `False`  and the user tries to set `init_n_iter`, also if `iter` < `init_n_iter`, etc.

PR #512, opened 2026-09-16
Rebranch the mixture model

KaisaridiSofia:
This PR is an attempt to try to rebranch the mixture model lower, ideally in the `LogisticModel`to facilitate future refactoring.


KaisaridiSofia:
Hello @MendezSebastianP . I managed to branch the mixture model to the logistic and I kept all the functions that differ. In my computer when I run the example from the user guide on the mixture model gives me exactly the same results.

I created a new file `mixture_new.py`. Once you give me a review I can delete the old mixture model and clean up the code in factory, init, e.t.c.

MendezSebastianP:
Very good job ! overall it is a clear improvement, now the code is cleaner and the model is placed in a better spot, i will post little by litte some comments with things to improve. These 2 points are related:
* First, you replaced `LogisticMultivariateMixtureModel` with `MixtureModel`, however, when i search `LogisticMultivariateMixtureModel`  in the repo i got 12 results, this could be problematic in many ways, as either the model is not fully adapted or there is some dead code in there.
* The file should be name `mixture.py,` not `mixture_new.py`, i need you to remove the old file and old code and just keep the functional code. This will also allow you to check what works and what do not.  Do not forget to check test/unit_tests/models/test_mixture_model.py (change just the names of the model inside), also `plotting.py` and `factory.py`, src/leaspy/models/__init__.py

MendezSebastianP:
Could you quote reply this comment and tell me what tests you did to make sure that the code still fully works? compared with the previous version.

KaisaridiSofia:
> Could you quote reply this comment and tell me what tests you did to make sure that the code still fully works? compared with the previous version.

I ran the mixture example on the examples of the user guide and got the same results.

MendezSebastianP:
> > Could you quote reply this comment and tell me what tests you did to make sure that the code still fully works? compared with the previous version.
> 
> I ran the mixture example on the examples of the user guide and got the same results.

Thats pretty good! could you also run your other scripts and also try to generate visualizations ? after that could you document here what you tested? being clear about this matter could allow future devs/phds to make reproductible studies, even reproduce those you made. Sometimes a model can diverge after many iterations, it is like the issue we had with Intel and AMD, the differences started to be visible after iter 2000, it could take more.

KaisaridiSofia:
> > > Could you quote reply this comment and tell me what tests you did to make sure that the code still fully works? compared with the previous version.
> > 
> > 
> > I ran the mixture example on the examples of the user guide and got the same results.
> 
> Thats pretty good! could you also run your other scripts and also try to generate visualizations ? after that could you document here what you tested? being clear about this matter could allow future devs/phds to make reproductible studies, even reproduce those you made. Sometimes a model can diverge after many iterations, it is like the issue we had with Intel and AMD, the differences started to be visible after iter 2000, it could take more.

Ok I see. I am not sure I will manage to do that in the time that I have left. I can clean up the code and update the tests and let you test the new model with the people that are still using it in the lab to verify the results in depth.

MendezSebastianP:
> > > > Could you quote reply this comment and tell me what tests you did to make sure that the code still fully works? compared with the previous version.
> > > 
> > > 
> > > I ran the mixture example on the examples of the user guide and got the same results.
> > 
> > 
> > Thats pretty good! could you also run your other scripts and also try to generate visualizations ? after that could you document here what you tested? being clear about this matter could allow future devs/phds to make reproductible studies, even reproduce those you made. Sometimes a model can diverge after many iterations, it is like the issue we had with Intel and AMD, the differences started to be visible after iter 2000, it could take more.
> 
> Ok I see. I am not sure I will manage to do that in the time that I have left. I can clean up the code and update the tests and let you test the new model with the people that are still using it in the lab to verify the results in depth.

I understand, i could totally run the scripts and do the comparison if you are ok, could you send me that?

KaisaridiSofia:
> > > > > Could you quote reply this comment and tell me what tests you did to make sure that the code still fully works? compared with the previous version.
> > > > 
> > > > 
> > > > I ran the mixture example on the examples of the user guide and got the same results.
> > > 
> > > 
> > > Thats pretty good! could you also run your other scripts and also try to generate visualizations ? after that could you document here what you tested? being clear about this matter could allow future devs/phds to make reproductible studies, even reproduce those you made. Sometimes a model can diverge after many iterations, it is like the issue we had with Intel and AMD, the differences started to be visible after iter 2000, it could take more.
> > 
> > 
> > Ok I see. I am not sure I will manage to do that in the time that I have left. I can clean up the code and update the tests and let you test the new model with the people that are still using it in the lab to verify the results in depth.
> 
> I understand, i could totally run the scripts and do the comparison if you are ok, could you send me that?

Yes I can do that. Let me clean up the scripts and the databases and I can send them to you next week. And you can also check with Maylis that is using the mixture model now for her analyses

KaisaridiSofia:
> Very good job ! overall it is a clear improvement, now the code is cleaner and the model is placed in a better spot, i will post little by litte some comments with things to improve. These 2 points are related:
> 
> * First, you replaced `LogisticMultivariateMixtureModel` with `MixtureModel`, however, when i search `LogisticMultivariateMixtureModel`  in the repo i got 12 results, this could be problematic in many ways, as either the model is not fully adapted or there is some dead code in there.
> * The file should be name `mixture.py,` not `mixture_new.py`, i need you to remove the old file and old code and just keep the functional code. This will also allow you to check what works and what do not.  Do not forget to check test/unit_tests/models/test_mixture_model.py (change just the names of the model inside), also `plotting.py` and `factory.py`, src/leaspy/models/**init**.py

I did that and subsequently I had to fix some other things to make it pass all the tests. I also cleaned up all the dead code. Let me know if now it's ok.

MendezSebastianP:
I just though about someting, if an user updates to v2.2  and launches `LogisticMultivariateMixtureModel`, the code will fail because it doesnt longer exist. This type of changes are reserved only for major releases, all code should keep working as it is. However, the change of name is the right choise so i added a few lines so when you call the old model it will run the new one as nothing happened

MendezSebastianP:
One other thing that i just spotted, in this case we need also a deprecation message, so users will migrate little by little to `MixtureModel` before we remove completetely `LogisticMultivariateMixtureModel` in the v3.0. next commit will fix this.

MendezSebastianP:
> I did that and subsequently I had to fix some other things to make it pass all the tests. I also cleaned up all the dead code. Let me know if now it's ok.
pretty nice job! it is almost done. I spotted one more thing, when testing for a `source_dimension=0` it crashed, idk if that happened before or not. Is this normal ? as `LogisticModel` allows the user to create a model with 0 sources. I will keep testing monday, normally if there is something left it should be small.

KaisaridiSofia:
Hi! Sorry for the last minute, but I noticed that the probabilities are not plotted in the convergence graph, so we need to add some code. Sorry I don't have time to do it now, and I cannot push directly but if you can just add this part it would be great.

In the `def __init__` function of the `MixtureModel` we need to add below this line:

`super().__init__(self.type, **kwargs)`

just that one:
```
self._configure_variables_to_track()
```

and then the function:

```
def _configure_variables_to_track(self):
            self.track_variables(["probs"])
```

and normally the convergence graphs should be as before

MendezSebastianP:
> Hi! Sorry for the last minute, but I noticed that the probabilities are not plotted in the convergence graph, so we need to add some code. Sorry I don't have time to do it now, and I cannot push directly but if you can just add this part it would be great.
> 
> In the `def __init__` function of the `MixtureModel` we need to add below this line:
> 
> `super().__init__(self.type, **kwargs)`
> 
> just that one:
> 
> ```
> self._configure_variables_to_track()
> ```
> 
> and then the function:
> 
> ```
> def _configure_variables_to_track(self):
>             self.track_variables(["probs"])
> ```
> 
> and normally the convergence graphs should be as before

No problem, i will take care of it:)

